PR #165, opened 2025-04-11
feat: Add a Collection Trait and implement for Array (#164)

### LINKED ISSUE #164: Add a Collection Trait and Implement It for Array
Create a `Collection` trait (NeoHaskell’s version of a type class) that encompasses all the functionality you’d typically expect from a collection. The design can draw inspiration from Swift’s [Collection protocol](https://developer.apple.com/documentation/swift/collection).

**Background / Motivation**  
- NeoHaskell currently has multiple collection types (e.g., `Array`, `Map`, etc.), each with its own set of functions.  
- Having a unified `Collection` trait would standardize and simplify working with these types.  
- This could make the codebase more coherent and make it easier for new contributors to understand collection-related operations.

**Proposed Plan**
1. **Define the `Collection` Trait**  
   - List out all the methods a typical collection should support (e.g., iteration, indexing, size/length, etc.).  
   - Look at Swift’s Collection protocol for inspiration, as well as existing NeoHaskell modules.

2. **Method Naming Convention**  
   - All methods in the `Collection` trait should be suffixed with `Impl` (e.g., `mapImpl`, `appendImpl`, etc.).  
   - These methods are not meant to be used directly.
   - This naming ensures that the exported functions in the specific collection modules (e.g., `Array.map`) are their monomorphic, public versions.

3. **Module Structure and Exports**  
   - Each collection module (e.g., `Array`, `Map`) continues exporting its own public API (`Array.map`, `Array.append`, etc.).
   - The `Collection` module also exports functions like `Collection.map`, which are wrappers over the corresponding `mapImpl` trait methods.
   - This allows both `Array.map` and `Collection.map` to be used in qualified form depending on the use case, while keeping typeclass machinery behind the scenes.

4. **Initial Implementation**  
   - Start with an implementation for `Array`, using existing functions in the module as a base.

5. **Extend Implementation**  
   - Gradually implement the trait for other core types such as `Map`, etc.

**Benefits**
- Improved consistency and abstraction over collection types.  
- Clearer and more modular API usage.  
- Easier onboarding for contributors.

**Next Steps**
1. Draft the `Collection` trait with the `Impl`-suffixed method names.
2. Define public wrappers in the `Collection` module.
3. Implement and test for `Array`. (Use doctests for this)
4. Extend to additional collection types.

ussgarci:
closes #164 

## Summary by CodeRabbit

* **New Features**
  * Introduced a new typeclass for collections, providing a unified interface for sequence-like data structures with a comprehensive set of operations.
  * Added a new function to retrieve the indices of an array.
  * Implemented an "isEven" utility for checking if a number is even.

* **Documentation**
  * Expanded and improved documentation and usage examples for many array and collection operations.

* **Bug Fixes**
  * Corrected the argument order in the array zipping functionality to ensure proper pairing of elements.

* **Chores**
  * Made the new collection module available for use by exposing it in the package configuration.


core/traits/Collection.hs
```diff
@@ -0,0 +1,44 @@
...
+appendImpl :: (Collection collection) -> (Collection collection) -> (Collection collection)
+appendImpl = Prelude.undefined
+
+
+firstImpl :: (Collection collection) -> Maybe b
+firstImpl = Prelude.undefined
+
+
+lastImpl :: (Collection collection) -> Maybe b
+lastImpl = Prelude.undefined
```
NickSeagull: Good start!

Actually for traits, we use Haskell's type class mechanism for now (until we get proper syntax for traits) so this would be defined like this actually:

```suggestion
-- | The Collection typeclass defines the behavior of a sequence whose
-- elements can be traversed and accessed by an indexed subscript.
class Collection collection where
  type Item collection

  countImpl   :: collection -> Int
  emptyImpl   :: collection
  isEmptyImpl :: collection -> Bool
  lengthImpl  :: collection -> Int
  getImpl     :: collection -> Int -> Maybe (Item collection)
  appendImpl  :: collection -> collection -> collection
  firstImpl   :: collection -> Maybe (Item collection)
  lastImpl    :: collection -> Maybe (Item collection)

-- | Wrapper functions without the Impl suffix
count :: Collection c => c -> Int
count = countImpl

empty :: Collection c => c
empty = emptyImpl

isEmpty :: Collection c => c -> Bool
isEmpty = isEmptyImpl

length :: Collection c => c -> Int
length = lengthImpl

get :: Collection c => c -> Int -> Maybe (Item c)
get = getImpl

append :: Collection c => c -> c -> c
append = appendImpl

first :: Collection c => c -> Maybe (Item c)
first = firstImpl

last :: Collection c => c -> Maybe (Item c)
last = lastImpl
```

This probably requires the language extension `TypeFamilies`. So you'd have to add that to `nhcore.cabal` along with the rest of extensions

ussgarci: Thanks Nick!  I will make that change and add the extension.

NickSeagull: You can actually click on "Commit suggestion" and then just pull locally for applying the suggestion i made :)

core/traits/Collection.hs
```diff
@@ -0,0 +1,65 @@
...
+
+
+  countImpl :: collection -> Prelude.Int
+  emptyImpl :: collection
+  isEmptyImpl :: collection -> Prelude.Bool
+  lengthImpl :: collection -> Prelude.Int
+  getImpl :: collection -> Prelude.Int -> Maybe (Item collection)
+  appendImpl :: collection -> collection -> collection
+  firstImpl :: collection -> Maybe (Item collection)
+  lastImpl :: collection -> Maybe (Item collection)
```
NickSeagull: You can actually remove the import of Prelude and just do `import Basics`

core/core/Array.hs
```diff
@@ -47,6 +47,7 @@ module Array (
 ) where
 
 import Basics
+import Collection qualified as CL
```
NickSeagull: This should be

```suggestion
import Collection qualified
import Collection (Collection(..))
```

core/core/Array.hs
```diff
@@ -67,6 +68,16 @@ newtype Array a = Array (Data.Vector.Vector a)
...
 
 
+instance CL.Collection (Array a) where
+  type Item (Array a) = a
+
+
+  countImpl (Collection c) =
+    (unwrap . CL.unwrap)
+      .> Data.Vector.length
+      .> Prelude.fromIntegral
```
NickSeagull: Also, instead of implementing the methods here, it's much better if we just wire them to the actual methods

```suggestion
  countImpl = length
```

Haven't checked this on my computer, but it should work properly

core/traits/Collection.hs
```diff
@@ -0,0 +1,72 @@
...
+first = firstImpl
+
+
+last :: (Collection c) => c -> Maybe (Collection.Item c)
+last = lastImpl
+
+
+-- | Helper function to unwrap a Collection
+unwrap :: Collection a -> a
+unwrap (Collection c) = c
```
NickSeagull: This one actually should go away, like, how do you unwrap an array? There are multiple elements, so there's no way to unwrap one

```suggestion
```

core/core/Array.hs
```diff
@@ -151,6 +158,13 @@ length :: Array a -> Int
 length = CL.lengthImpl
 
 
+-- | The indices that are valid for subscripting the collection, in ascending order.
+--
+-- > indices (fromLinkedList [1,2,3]) == [0,1,2]
+indices :: Array a -> [Int]
+indices = CL.indicesImpl
```
NickSeagull: Could you please flip the way you're doing this? I mean, implement the indices function here, and just do the linkage in `indicesImpl`? I know this is counterintuitive because of the Impl suffix, but it will make sense when I introduce traits formally and those will be just codegen'd

ussgarci: You got it.  Thanks for the feedback!

core/nhcore.cabal
```diff
@@ -162,4 +162,12 @@ test-suite nhcore-test
...
     build-depends:
         base ^>=4.18,
+        doctest ^>=0.24.0,
+        QuickCheck ,
+        vector ,
+        Basics,
+        Collection,
+        LinkedList,
+        Maybe,
+        Tuple,
```
NickSeagull: These are modules, won't work here, as this expects packages,you should add nhcore here. Or even better, rebase the branch to `feat/in-memory-event-store`

I can help with this if you want

ussgarci: Thanks @NickSeagull!  I will rebase today and reach out if I bungle it.  :slightly_smiling_face:

core/core/Array.hs
```diff
@@ -101,33 +142,53 @@ unwrap (Array v) = v
...
 
 -- | Determine if an array is empty.
 --
--- > isEmpty empty == True
+-- >>> isEmpty empty
+-- True
+-- >>> isEmpty (Array (Data.Vector.singleton 'a'))
+-- False
+-- >>> isEmpty (Array (Data.Vector.fromList [1,2,3] :: Data.Vector.Vector Int))
+-- False
```
NickSeagull: Can we make these like:

```suggestion
-- >>> isEmpty (Array.fromLinkedList ['a'])
-- False
-- >>> isEmpty (Array.fromLinkedList ['a', 'b', 'c']))
-- False
```

core/core/Array.hs
```diff
@@ -101,33 +142,53 @@ unwrap (Array v) = v
...
+length = unwrap .> Data.Vector.length
+
+
+-- | The indices that are valid for subscripting the collection, in ascending order.
+--
+-- >>> indices (Array (Data.Vector.fromList [1,2,3] :: Data.Vector.Vector Int))
+-- [0,1,2]
+-- >>> indices empty
+-- []
+indices :: Array a -> [Int]
```
NickSeagull: Can you return an Array of Int here? In NeoHaskell Arrays are preferred over lists. In fact in the future the `[something]` type will be forbidden at parser level

core/core/Array.hs
```diff
@@ -138,9 +199,11 @@ initialize n f =
 
 -- | Creates an array with a given length, filled with a default element.
 --
--- > repeat 5 0     == fromLinkedList [0,0,0,0,0]
--- > repeat 3 "cat" == fromLinkedList ["cat","cat","cat"]
---
+-- >>> repeat 5 (0 :: Int)
+-- Array [0,0,0,0,0]
+-- >>> repeat 3 ("cat" :: Prelude.String)
```
NickSeagull: This should be `Text`.


```suggestion
-- >>> repeat 3 ("cat" :: Text)
```

core/core/Array.hs
```diff
@@ -161,20 +226,36 @@ fromLinkedList =
...
+
+
+-- | Return @Just@ the first element or @Nothing@ if the index is empty.
+--
+-- >>> first (fromLinkedList [0,1,2] :: Array Int)
+-- Just 0
+-- >>> first (fromLinkedList [] :: Array Int)
+-- Nothing
+first :: Array a -> Maybe a
+first = unwrap .> (!? 0)
```
NickSeagull: I'd say this should be `get 0` so we dont duplicate logic

core/core/Array.hs
```diff
@@ -214,54 +298,61 @@ toIndexedLinkedList =
...
+-- 15
 reduce :: (a -> b -> b) -> b -> Array a -> b
 reduce f value array = Prelude.foldr f value (unwrap array)
 
 
 -- | Reduce an array from the left. Read @foldl@ as fold from the left.
 --
--- > foldl (:) [] (fromLinkedList [1,2,3]) == [3,2,1]
+-- >>> foldl (:) [] (fromLinkedList [1,2,3] :: Array Int)
+-- [3,2,1]
```
NickSeagull: Maybe something like
```suggestion
-- >>> (fromLinkedList [1,2,3] :: Array Int) |> foldl append empty 
-- [3,2,1]
```

ussgarci:
Thanks @NickSeagull!  I will make the suggested changes.

core/core/Array.hs
```diff
@@ -101,33 +142,53 @@ unwrap (Array v) = v
...
+-- Array [0,1,2]
+-- >>> indices empty
+-- Array []
+indices :: Array a -> Array Int
+indices =
   unwrap
-    .> Data.Vector.length
-    .> Prelude.fromIntegral
+    .> \v ->
+      Array (Data.Vector.generate (Data.Vector.length v) Prelude.id)
```
NickSeagull: Instead of `id` one can use `unchanged` in `Basics`

```suggestion
      Array (Data.Vector.generate (Data.Vector.length v) unchanged)
```

NickSeagull (commented):
looks good! Left one nitpick comment. 

Can you take a look at the failing tests? Yes they are related to event store, but because those tests make extensive use of the Array module. Let me know if you need help with that.

Hopefully by updating the branch they will fix themselves, but if not, just ping me and I'll lend you a hand!

core/testlib/Test/Service/EventStore/GlobalStreamOrdering/Spec.hs
```diff
@@ -75,9 +75,7 @@ spec newStore = do
           context.store.readAllEventsForwardFrom (Event.StreamPosition 0) (EventStore.Limit (expectedTotalEvents))
             |> Task.mapError toText
         Task.unless ((Array.length allGlobalEvents) <= 1) do
-          let eventPairs =
-                allGlobalEvents
-                  |> Array.zip (Array.drop 1 allGlobalEvents)
+          let eventPairs = Array.zip allGlobalEvents (Array.drop 1 allGlobalEvents)
```
NickSeagull: Can you modify `Array.zip` so this modification is not needed? The idea is that we dont touch the tests.

Remember that in NeoHaskell, the Array argument (this/self in other languages) must go **last**, contrary to the regular functions in Haskell.

Once this is solved, I'll merge this :)

ussgarci: Done! :)

NickSeagull:
@all-contributors please add @ussgarci for code

PR #446, opened 2026-03-06
fix(core): prevent LISTEN/NOTIFY connection leak in test teardown

### LINKED ISSUE #439: CI: LISTEN/NOTIFY connections leak during test teardown, exhausting PostgreSQL max_connections
## Problem

`nhcore-test-service` fails on macOS CI with 25 out of 566 tests failing due to PostgreSQL connection exhaustion:

```
FATAL: sorry, too many clients already
```

Failing run: https://github.com/neohaskell/NeoHaskell/actions/runs/22721463607/job/65884488244

## Root Cause

Test teardown drops tables (including `Events`) but does not terminate the LISTEN/NOTIFY background threads spawned by `Notifications.connectTo`. The orphaned listeners:

1. Detect the dropped table and log `"LISTEN/NOTIFY listener exited unexpectedly"` (see #405 — this fires unconditionally due to the `AsyncTask.run`/`Task.andThen` bug)
2. Keep attempting to query the now-missing `Events` table, producing hundreds of `relation "events" does not exist` warnings
3. Each reconnect attempt consumes a new PostgreSQL connection
4. By the time later test specs run (`EntityFetcher.Fetch.Spec`, `CommandHandler.Execute.Spec`, `FileUpload.FileStateStore`), `max_connections` is exhausted

## Affected Tests

All 25 failures share the same error — `SubscriptionInitializationError` or `Failed to initialize file state table` with `"sorry, too many clients already"`:

- `Test.Service.EntityFetcher.Fetch.Spec` — 2 failures
- `Test.Service.CommandHandler.Execute.Spec` — 11 failures
- `Test.Service.FileUpload.FileStateStore` — 12 failures

## Suggested Fix

1. Add a `close`/`destroy` function to the EventStore (or its Postgres backend) that terminates the LISTEN/NOTIFY async thread and releases its connection
2. Call that cleanup in test teardown hooks (`afterAll`/`after`) before dropping tables
3. Optionally, add a CI safety net that runs `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE ...` between test spec groups

## Related Issues

- #397 — LISTEN connection has no keepalive or reconnection (the underlying lack of connection lifecycle management)
- #405 — False "exited unexpectedly" log on every startup (contributes to the reconnect spam pattern)

NickSeagull:
Closes #439

## Root Cause

`nhcore-test-service` was failing on macOS CI with 25/566 tests failing due to PostgreSQL connection exhaustion (`FATAL: sorry, too many clients already`). The root causes were:

1. **Fire-and-forget listener**: `Notifications.connectTo` called `AsyncTask.run |> discard` — handle thrown away, impossible to cancel
2. **Connection leak per retry**: Each reconnect acquired 2 raw connections never released
3. **Infinite reconnect on table drop**: Tests drop Events table → listener crashes → reconnects infinitely
4. **No test teardown**: Test framework had `afterAll`/`after` hooks commented out
5. **Orphaned initial connection**: `initializeSubscriptions` leaked `initialListenConnection`

## Changes

- **`Notifications.hs`**: `connectTo` now returns `Task Text (Task Text Unit)` — the inner Task is a cleanup action that cancels the listener and releases connections
- **`Core.hs`**: `EventStore` record gains `close :: Task Text Unit` field for resource lifecycle management
- **`Internal.hs`**: `initializeSubscriptions` returns cleanup action; `initialListenConnection` released after trigger setup; `EventStore.close` stops listener + releases pool
- **`Simple.hs`**: No-op `close` implementation
- **`Test/Spec.hs`**: `after`/`afterAll` hooks enabled; `beforeAll`/`before` auto-cleanup via `AutoCleanup` typeclass
- **`PostgresSpec.hs`**: `newStore`/`newCartStore` call `store.close` before dropping tables
- **`Subscriptions/Spec.hs`**: `Task.finally` wrapping for guaranteed unsubscribe cleanup
- **`FileStateStore/PostgresSpec.hs`**: Explicit pool release after `dropTable`
- **`NotificationsSpec.hs`**: Regression test for listener cleanup

## Verification

- `cabal build nhcore-test-service` ✅ (30/30 modules)
- `cabal test nhcore-test-service` ✅ (572 examples, 0 failures, 2 pending)
- ADR-0039 documents the architectural decision

## Summary by CodeRabbit

* **New Features**
  * EventStore instances now expose an explicit close operation for controlled resource cleanup.
  * Notification/listener setup now returns a cleanup handle to stop background listeners.

* **Bug Fixes**
  * Fixed LISTEN/NOTIFY connection leak in PostgreSQL stores, preventing reconnection storms and CI failures.

* **Tests**
  * Added regression tests for listener shutdown and store.close behavior.
  * Test framework gained automatic after/afterAll cleanup hooks to ensure reliable teardown.

* **Documentation**
  * Added ADR documenting the connection-leak fix and lifecycle changes.

### DIFF
--- a/core/service/Service/EventStore/Postgres/Internal.hs
+++ b/core/service/Service/EventStore/Postgres/Internal.hs
@@ -159,8 +159,12 @@ defaultOps = do
               |> Task.mapError toText
             let connectionFactory = do
                   listenConnection <- Hasql.acquire (toConnectionSettings cfg) |> Task.fromIOEither |> Task.mapError toText
-                  queryConnection <- Hasql.acquire (toConnectionSettings cfg) |> Task.fromIOEither |> Task.mapError toText
-                  Task.yield (listenConnection, queryConnection)
+                  queryResult <- Hasql.acquire (toConnectionSettings cfg) |> Task.fromIOEither |> Task.mapError toText |> Task.asResult
+                  case queryResult of
+                    Ok queryConnection -> Task.yield (listenConnection, queryConnection)
+                    Err err -> do
+                      Hasql.release listenConnection |> Task.fromIO
+                      Task.throw err
             subscriptionStore |> Notifications.connectTo connectionFactory
 
   let release connection = do
@@ -229,9 +233,10 @@ new ops cfg = do
               subscribeToStreamEvents = subscribeToStreamEventsImpl ops cfg subscriptionStore,
               unsubscribe = unsubscribeImpl subscriptionStore,
               truncateStream = truncateStreamImpl ops cfg,
-              close = do
-                cleanup
-                ops.release pool |> Task.mapError toText |> discard
+              close =
+                Task.finally
+                  (ops.release pool |> Task.mapError toText |> discard)
+                  cleanup
             }
     Task.yield eventStore
   case result of
--- a/core/service/Service/EventStore/Postgres/Notifications.hs
+++ b/core/service/Service/EventStore/Postgres/Notifications.hs
@@ -35,21 +35,22 @@ connectTo acquireConnections store = do
           result <- Task.asResultSafe do
             (listenConnection, queryConnection) <- acquireConnections
             let releaseConns = do
+                  currentConnectionsRef |> Var.set Maybe.Nothing
                   Hasql.release listenConnection |> Task.fromIO
                   Hasql.release queryConnection |> Task.fromIO
-            trackResult <- Task.asResultSafe (currentConnectionsRef |> Var.set (Maybe.Just (listenConnection, queryConnection)) :: Task Text Unit)
-            case trackResult of
-              Err _ -> releaseConns
-              Ok _ -> pass
-            let channelToListen = HasqlNotifications.toPgIdentifier "global"
-            HasqlNotifications.listen listenConnection channelToListen
-              |> Task.fromIO
-              |> discard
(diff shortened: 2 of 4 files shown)


NickSeagull:
All 6 CodeRabbit review comments addressed in commit dae9147:

1. **Internal.hs double-release**: `new` restructured to own the pool directly (no longer uses `withConnection`). Pool is released exactly once — either in `close` on success, or in the error handler on init failure.

2. **Internal.hs initializeSubscriptions exception safety**: Wrapped session runs in `Task.finally` so `initialListenConnection` is always released even if a session throws.

3. **Notifications.hs race condition**: After `acquireConnections`, wrapped `Var.set` in `Task.asResultSafe` and explicitly release connections if tracking fails (guards against async exception between acquire and set).

4. **Test/Spec.hs OVERLAPPABLE fallback**: Removed the blanket `{-# OVERLAPPABLE #-}` instance. Added explicit no-op `AutoCleanup` instances for the 3 context types that don't need cleanup (`CommandDecideContext`, `QueryObjectStoreInMemoryContext`, `SnapshotCacheInMemoryContext`).

5. **ADR-0039 count**: "four compounding problems" → "five compounding problems".

6. **ADR-0039 path**: Fixed `Notifications.hs` link to include `Postgres/` segment.

PR #634, opened 2026-05-14
feat(TH): fold JSON + deriving boilerplate into concept markers (#630)

### LINKED ISSUE #630: feat(TH): consolidate JSON + deriving boilerplate into command/query/event markers
## Summary

Today, a NeoHaskell command/query/event file has three layers of boilerplate around the actual data declaration:

1. A `deriving (Generic, Typeable, Show)` clause on the data type.
2. Two empty instance declarations: `instance Json.FromJSON X` and `instance Json.ToJSON X`.
3. The concept-marker TH call (`command ''X`, `deriveQuery ''X [''Entity]`).

The TH markers already know the type at the moment they run; they could generate (1) and (2) themselves via standalone deriving, leaving the user with just *data + functions + marker*.

## Use Case

A real consumer command (evoca-payment-gateway, `ConfirmDonationDeclined.hs`) looks like this:

```haskell
data ConfirmDonationDeclined = ConfirmDonationDeclined
  { donationId :: Uuid
  , declineReason :: Text
  }
  deriving (Generic, Typeable, Show)


instance Json.FromJSON ConfirmDonationDeclined


instance Json.ToJSON ConfirmDonationDeclined


-- ... getEntityId, decide, type instances ...


command ''ConfirmDonationDeclined
```

The first 8 lines are essentially ceremony — the same shape on every command, every event, every query, with zero per-type variation. An event file is the same minus `Typeable`. A query adds an empty `ToSchema` instance on top.

## Proposed API

Have `command`, `deriveQuery`, and a new `event` TH function generate:

- `deriving stock instance Show X`
- `deriving stock instance Generic X`
- `deriving stock instance Typeable X` (commands only — events don't currently need it)
- `instance Json.FromJSON X`
- `instance Json.ToJSON X`
- For queries: `instance ToSchema X` (or whatever the schema default is)

…using `StandaloneDeriving` so the data declaration stays clean. After the change, the consumer file looks like:

```haskell
data ConfirmDonationDeclined = ConfirmDonationDeclined
  { donationId :: Uuid
  , declineReason :: Text
  }


-- ... getEntityId, decide, type instances ...


command ''ConfirmDonationDeclined
```

…and the marker quietly emits everything underneath. The `event` marker (currently missing) would do the same job for event types, which today still need the FromJSON/ToJSON/Generic/Show ceremony hand-written.

For escape hatches: if a user needs a non-default JSON encoding (e.g. tagged sum), they can still write the instance explicitly — the TH should be a no-op for any class instance the user has already defined.

## Alternatives Considered

- **Status quo (per-file ceremony).** Every new command/event/query repeats the same 5–6 lines verbatim, with all the failure modes that brings: missed instances cause runtime JSON parse errors, missed `Generic` breaks the TH that depends on it, etc.
- **A generic Template-Haskell helper (e.g. `deriveJsonAndGeneric ''X`) that's not tied to command/query/event.** Less ergonomic — the user still has to remember to call it on every type. Folding it into the existing concept markers makes it impossible to forget.
- **CPP / preprocessor.** Hard no on the NeoHaskell side; this is exactly what TH is for.

## Notes

A related concern: today the data declaration has `Generic, Typeable, Show` derived inline. If the TH later wants to add a `deriving stock` standalone clause for `Generic`, there'd be a duplicate-instance error. The TH should either:
(a) Strip those derives via reify + re-emit (more involved), or
(b) Document that consumers should *not* include `deriving (Generic, Typeable, Show)` once the new markers ship, and provide a migration codemod.

Spotted while landing slices in a real consumer project (evoca-payment-gateway). Happy to send a PR if the API shape is acceptable.

NickSeagull:
## Summary

Before this PR, writing a new command in NeoHaskell looked like this:

```haskell
data CreateCart = CreateCart
  deriving (Generic, Typeable, Show)


instance Json.FromJSON CreateCart


instance Json.ToJSON CreateCart


-- ... getEntityId, decide, type instances ...


command ''CreateCart
```

After this PR, it looks like this:

```haskell
data CreateCart = CreateCart


-- ... getEntityId, decide, type instances ...


command ''CreateCart
```

Five fewer lines — every time, for every command, event, and query. The `command`, `deriveQuery`, and new `event` TH markers now emit `Show`, `Generic`, `FromJSON`, and `ToJSON` (plus `ToSchema` for queries) idempotently: each marker calls `TH.reifyInstances` at splice time and only emits a declaration if no instance is already in scope. Existing files keep compiling unchanged — the markers no-op when the instance is already there — so there is no migration. Custom JSON encodings (tagged sums, `omitNothingFields`) work the same as before: declare the instance *before* the marker call and the marker skips it.

ADR-0057 classifies this as a **moderate**-tier change: extends the `command`/`deriveQuery` TH functions and adds a new `event` function, no runtime IO, no new dependency, no secrets or auth surface.

## ADR

`docs/decisions/0057-th-marker-boilerplate.md` (Status: Proposed → lands with this PR).

Key design choices documented there:
- **Idempotent TH chosen** over stripping inline `deriving` (TH cannot rewrite an already-spliced data declaration) and over a standalone `deriveBoilerplate ''X` helper (forgettable; folding into the concept markers makes it impossible to forget).
- **`reifyInstances`** is the mechanism — a hash-map read inside GHC's instance environment, negligible cost at splice time.
- **Marker-last convention is now load-bearing**: declaring a custom instance *after* the marker call would produce a duplicate-instance compile error. This is documented in the ADR, the module header of `Service.Event.TH`, and flagged as a medium finding (TM-001) to be documented in `AGENTS.md`.

## Changes

**New files:**
- `core/service/Service/CommandExecutor/TH/Internal.hs` — `emitInstanceIfMissing :: TH.Name -> TH.Name -> THLib.DecsQ -> THLib.DecsQ`, the idempotency primitive shared by all three markers.
- `core/service/Service/Event/TH.hs` — new `event :: TH.Name -> THLib.DecsQ` marker.
- `docs/decisions/0057-th-marker-boilerplate.md` — ADR.
- `core/test/Service/CommandExecutor/THSpec.hs` + 4 fixture files under `core/test/Service/CommandExecutor/TH/` — tests for the extended `command` behaviour.
- `core/test/Service/Event/THSpec.hs` — tests for the new `event` marker.

**Modified files:**
- `core/service/Service/CommandExecutor/TH.hs` — `command` now calls `emitInstanceIfMissing` for `Show`/`Generic`/`FromJSON`/`ToJSON`.
- `core/service/Service/Query/TH.hs` — `deriveQuery` now calls `emitInstanceIfMissing` for `Show`/`Generic`/`FromJSON`/`ToJSON`/`ToSchema`.
- `core/test/Service/Query/THSpec.hs` — new cases for the extended `deriveQuery` behaviour.
- `core/test-service/Main.hs` — registers the two new Spec modules.
- `core/nhcore.cabal`, `testbed/nhtestbed.cabal` — add `StandaloneDeriving` and `DerivingStrategies` to `default-extensions`; register new modules.

## Security

Two review passes (phases 4 and 12). **0 blockers.** 17 informational findings across both passes, all demoted by grounding:

- **17 `String`-type-in-TH flags** — `String` is the correct type for Template Haskell name manipulation; these modules are compile-time only and have no runtime attacker surface (fails Q1 blast-radius).
- **2 `Hashable.hash` determinism flags** — `Hashable.hash` is called at compile time for `KnownHash` instances; the output is a fixed integer compiled into the binary, no runtime determinism concern (fails Q1).
- **1 medium finding kept (TM-001)** — the "marker last" convention is now load-bearing; a custom instance declared *after* the marker call fails to compile with a duplicate-instance error. Mitigation: document the convention in `AGENTS.md` and the `Service.Event.TH` module header. Not a blocker; the error is a compile-time duplicate-instance message, not a silent misbehaviour.

## Performance

Two review passes (phases 5 and 13). **0 blockers.** 18 informational findings, all demoted by grounding:

- All `String`-type and `[fmt|...|]`-allocation flags are on compile-time-only TH code; they do not touch the 50k req/s runtime hot path (fails Q2 reachability).
- The `reifyInstances` cost at splice time is non-zero but negligible (hash-map reads inside GHC); documented in the ADR Consequences.
- Generic-defaulted `ToJSON` is not forced on anyone — the escape hatch (custom instance before the marker) preserves hand-written `toEncoding` for hot paths (fails Q1 blast-radius).

## Tests

All 7 suites pass: **1604 examples, 6 pending, 0 failures.**

| Suite | Result |
|---|---|
| nhcore-test | pass |
| nhcore-test-core | pass |
| nhcore-test-auth | pass |
| nhcore-test-integration | pass |
| nhcore-test-service | pass |
| nhintegrations-test | pass |
| neohaskell-lsp-test | pass |

New test coverage added:
- Happy path: `command`/`event`/`deriveQuery` emit all four (five for queries) instances when none are declared.
- Idempotency: markers emit nothing when all instances are already in scope.
- Escape hatch: custom `ToJSON` declared before the marker; marker emits only the missing instances.
- Boundary: partial pre-declaration (e.g. only `FromJSON` present); marker fills in the rest.

## Hlint

No hints on changed files.

## Checklist

- [x] ADR approved (phase 3 approved)
- [x] Security findings grounded (phase 12 complete — 0 blockers, 17 informational)
- [x] Performance findings grounded (phase 13 complete — 0 blockers, 18 informational)
- [x] Tests passing (1604 examples, 0 failures across 7 suites)
- Hlint: clean

Closes #630

## Summary by CodeRabbit

* **New Features**
  * Template Haskell markers (`command`, `deriveQuery`) now automatically generate `Show`, `Generic`, `FromJSON`, and `ToJSON` instances.
  * New `event` Template Haskell marker available for streamlined event type definitions.

* **Documentation**
  * Added architecture decision record detailing Template Haskell marker enhancements and migration guidance.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,24 @@ When unavoidable, prefix with `Ghc`:
 import Data.List qualified as GhcList
 ```
 
+### Concept Markers (`command`, `event`, `deriveQuery`)
+
+Each marker emits `Show`, `Generic`, `Json.FromJSON`, and `Json.ToJSON`
+(plus `ToSchema` for queries) for the named type — idempotently, via
+`TH.reifyInstances`. Two rules:
+
+- **Marker last.** The marker splice must come *after* every other
+  declaration in the file, including any custom `Json.ToJSON` /
+  `Json.FromJSON` instance you want to keep. Declaring a custom instance
+  *after* the marker produces a `Duplicate instance declarations`
+  compile error because the marker has already emitted the derived one.
+- **Custom encodings stay possible.** Want a tagged-sum `Json.ToJSON`
+  or `omitNothingFields`? Declare it *before* the marker call; the
+  marker detects it via `reifyInstances` and emits no duplicate.
+
+The four-pack helper lives at `Service.TH.Boilerplate` and is shared by
+all three markers — see ADR-0057.
+
 ## ANTI-PATTERNS (FORBIDDEN)
 
 - `let..in` or `where` clauses
--- a/core/nhcore.cabal
+++ b/core/nhcore.cabal
@@ -100,6 +100,7 @@ common common_cfg
     BlockArguments
     DataKinds
     DeriveDataTypeable
+    DerivingStrategies
     DuplicateRecordFields
     ImportQualifiedPost
     ImplicitParams
@@ -115,6 +116,7 @@ common common_cfg
     PartialTypeSignatures
     QualifiedDo
     QuasiQuotes
+    StandaloneDeriving
     Strict
     TemplateHaskell
     TypeApplications
@@ -223,12 +225,14 @@ library
     Service.CommandExecutor
     Service.CommandExecutor.Core
     Service.CommandExecutor.TH
+    Service.CommandExecutor.TH.Internal
     Service.Entity
     Service.Entity.Core
     Service.EntityFetcher
     Service.EntityFetcher.Core
     Service.Error
     Service.Event
+    Service.Event.TH
     Service.Event.EntityName
     Service.Event.EventMetadata
     Service.Event.StreamId
@@ -484,6 +488,12 @@ test-suite nhcore-test
     Service.Integration.SelectionSpec
     Service.Integration.TestFixtures
     Service.Integration.WireupSpec
+    Service.CommandExecutor.TH.AlreadyDerivedCommandFixture
+    Service.CommandExecutor.TH.FreshCommandFixture
+    Service.CommandExecutor.TH.PartialDerivedCommandFixture
+    Service.CommandExecutor.TH.SumCommandFixture
+    Service.CommandExecutor.THSpec
+    Service.Event.THSpec
 
   -- other-extensions:
   type: exitcode-stdio-1.0
@@ -535,8 +545,14 @@ test-suite nhcore-test-service
 
   other-modules:
     Service.ApplicationSpec
+    Service.CommandExecutor.THSpec
+    Service.CommandExecutor.TH.AlreadyDerivedCommandFixture
+    Service.CommandExecutor.TH.FreshCommandFixture
+    Service.CommandExecutor.TH.PartialDerivedCommandFixture
+    Service.CommandExecutor.TH.SumCommandFixture
     Service.CommandHandlerSpec
     Service.CommandSpec
+    Service.Event.THSpec
     Service.EventStore.InMemorySpec
     Service.EventStore.SimpleSpec
     Service.EventStore.Postgres.SubscriptionStoreSpec
--- a/core/nhcore.cabal
+++ b/core/nhcore.cabal
@@ -225,7 +225,6 @@ library
     Service.CommandExecutor
     Service.CommandExecutor.Core
     Service.CommandExecutor.TH
-    Service.CommandExecutor.TH.Internal
     Service.Entity
     Service.Entity.Core
     Service.EntityFetcher
@@ -283,6 +282,7 @@ library
     Service.FileUpload.Web
     Service.ServiceDefinition
     Service.ServiceDefinition.Core
+    Service.TH.Boilerplate
     Service.Transport
     Service.Transport.Web
     Service.Transport.Web.BuiltinSchemas
--- a/core/service/Service/CommandExecutor/TH.hs
+++ b/core/service/Service/CommandExecutor/TH.hs
@@ -5,12 +5,14 @@ module Service.CommandExecutor.TH (
 
 import Control.Monad.Fail qualified as MonadFail
 import Core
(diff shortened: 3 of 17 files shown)


NickSeagull:
Both CodeRabbit findings addressed in [`dd652cd`](https://github.com/neohaskell/NeoHaskell/pull/634/commits/dd652cd):

- ADR-0057 `Status` flipped from `Proposed` → `Accepted` (per the 55+ existing Accepted ADRs in the repo).
- AGENTS.md gained a `Concept Markers` subsection under CONVENTIONS covering the marker-last rule + the custom-encoding escape hatch, pointing at `Service.TH.Boilerplate`.

CI fully green (9/9 checks, build 5m30s, test-core 7m3s). Proceeding to merge per maintainer instruction.

PR #642, opened 2026-05-19
feat(service): per-command AuthOptions — let commands declare Everyone vs Authenticated (#638)

### LINKED ISSUE #638: feat: Per-command AuthOptions — let commands declare Everyone vs Authenticated
## Summary

NeoHaskell currently treats command auth as all-or-nothing: when `Application.withAuth` is wired, **every** command endpoint requires a valid JWT, regardless of which command. There is no way for an app to expose a single public command (e.g. anonymous form submission, newsletter signup, public-facing webhook receiver) alongside authenticated commands in the same service.

Queries already support per-endpoint auth via `Service.Query.Auth` (`publicAccess` vs `authenticatedAccess`, threaded through `canAccessImpl`). Commands should follow the same pattern.

## Current behavior

The `AuthOptions` ADT already exists with both modes:

```haskell
-- core/auth/Auth/Options.hs:12-16
data AuthOptions
  = Everyone        -- No authentication required
  | Authenticated   -- Valid JWT required
```

But the Web transport's command-dispatch path hardcodes `Authenticated`:

```haskell
-- core/service/Service/Transport/Web.hs:583
authResult <- Middleware.checkAuth (Just auth.jwksManager) auth.authConfig Authenticated request
```

Same hardcoding at lines 783 and 823. The *only* place `Everyone` is used at runtime is the file-download path (`Web.hs:876`), and it is hardcoded there too — not selectable per endpoint.

`Service.Command.Core` has no auth-related field or type family. `TransportsOf` controls *which transports* a command is exposed on (Web, Internal, CLI, MCP) but not its auth mode. There is no fallthrough analogous to how queries handle `TokenMissing` at `Web.hs:671-673`.

Net result: an app that wires `withAuth` to gate any command effectively gates all of them. The only escape hatch is to disable auth globally for the whole application (typically via an `AUTH_DISABLED=true` env var pattern in the launcher), which is not a real fix — it removes auth from every command, including the ones that genuinely need it, and leaves the architecture lying about its intent.

## Proposed API

Mirror the query pattern. Add a type family with a safe-by-default instance:

```haskell
-- core/service/Service/Command/Core.hs
type family AuthOptionsOf (commandType :: Type) :: AuthOptions

-- Default behavior is preserved: commands without an explicit instance
-- remain Authenticated, exactly like today.
```

Commands opt into public access explicitly:

```haskell
data SubmitFeedback = SubmitFeedback { message :: Text, ... }

type instance EntityOf SubmitFeedback = FeedbackEntity
type instance TransportsOf SubmitFeedback = '[WebTransport]
type instance AuthOptionsOf SubmitFeedback = 'Everyone   -- explicit opt-in
```

At the Web transport, replace the hardcoded literal with the per-command value resolved at TH/registration time:

```haskell
-- Web.hs:583 — after change
authResult <- Middleware.checkAuth
                (Just auth.jwksManager)
                auth.authConfig
                endpoint.authOptions    -- carried on EndpointHandler metadata
                request

case authResult of
  Result.Err authErr -> case endpoint.authOptions of
    Everyone      -> processCommandWithContext =<< Auth.anonymousContext
    Authenticated -> Middleware.respondWithAuthError authErr respond
  Result.Ok authContext -> ...
```

This matches the existing query semantics at `Web.hs:665-679`: a missing token on an `Everyone` endpoint falls through to anonymous context; an invalid token still rejects.

## Why a type family (matching `TransportsOf`)

`TransportsOf` is already a type family resolved at TH time and read by the command TH macro to build per-transport endpoint metadata. `AuthOptionsOf` slots in next to it with zero new mechanism — the TH macro at `Service/CommandExecutor/TH.hs` simply asks for one more type instance per command and threads the resolved value into the registered endpoint.

A value-level `canAccess` (the query approach) would also work, but the type-family approach keeps command registration uniform with `TransportsOf` and avoids requiring an extra top-level binding per command.

## Files that need changes

| File | Change |
|------|--------|
| `core/service/Service/Command/Core.hs` | Add `type family AuthOptionsOf` |
| `core/service/Service/CommandExecutor/TH.hs` | Look up `AuthOptionsOf` instance; default to `Authenticated` if absent; capture the resolved `AuthOptions` on the registered endpoint |
| `core/service/Service/Transport.hs` | Extend `EndpointHandler` (or sibling map) to carry the per-command `AuthOptions` |
| `core/service/Service/Transport/Web.hs` | Read `authOptions` from endpoint metadata at lines 583, 783, 823 instead of hardcoding `Authenticated`; on `TokenMissing` fall through to anonymous context for `Everyone` commands (mirror the query path at 671-673) |
| `core/service/Service/Transport/Mcp.hs` | Same treatment if MCP gates commands by auth |
| `core/service/Service/Transport/Cli.hs` | No change expected (CLI is not JWT-gated) |
| `core/test/Auth/MiddlewareSpec.hs` | Add cases for command-side `Everyone` (missing token → anonymous, invalid token → 401) |
| `core/test/Service/CommandExecutor/THSpec.hs` | Cover both "no `AuthOptionsOf` instance → defaults to `Authenticated`" and "explicit `'Everyone` instance" |

No changes needed to:

- `core/auth/Auth/Options.hs` — `AuthOptions` already has both constructors
- `core/auth/Auth/Middleware.hs` — `checkAuth` already takes `AuthOptions` as a parameter (lines 129-136); `Everyone` semantics already implemented
- Individual app code — commands without an `AuthOptionsOf` instance keep their current `Authenticated` behavior

## Backwards compatibility

The default for the type family is `Authenticated`. Apps that do not add an `AuthOptionsOf` instance behave exactly as they do today. The change is strictly additive: no existing command's auth surface changes unless the author opts in.

## Acceptance criteria

- [ ] `type family AuthOptionsOf :: Type -> AuthOptions` added to `Service.Command.Core` with a documented default of `Authenticated`
- [ ] Command TH macro resolves `AuthOptionsOf` for each registered command and threads the value into endpoint metadata; absent instance resolves to `Authenticated`
- [ ] `EndpointHandler` (or equivalent registration record) carries the resolved `AuthOptions`
- [ ] `Service.Transport.Web` command path passes the per-command `AuthOptions` to `Middleware.checkAuth` at the three sites that currently hardcode `Authenticated` (lines 583, 783, 823)
- [ ] On `Everyone` commands, a missing JWT (`TokenMissing`) is treated as anonymous and the command runs with `Auth.anonymousContext` — mirroring the existing query fallthrough at `Web.hs:671-673`
- [ ] On `Everyone` commands, an *invalid* JWT still returns 401 (do not silently accept a malformed token)
- [ ] Commands with `AuthOptionsOf = 'Authenticated` (explicit or default) keep returning 401 on missing/invalid tokens (no regression)
- [ ] `OAuth2` and other command-bearing endpoints in `Web.hs` (lines 783, 823) honor the per-command setting consistently
- [ ] Tests covering both the type-family resolution (TH) and runtime auth dispatch (middleware + transport)
- [ ] Documentation in `Service.Command.Core` and a short example mirroring the query-auth docstring style ("default to authenticated, public access is explicit opt-in")

NickSeagull:
# Per-command AuthOptions — let commands declare Everyone vs Authenticated

## Summary

This PR implements ADR-0058, allowing individual commands to declare their own authentication requirements via a `canAccess` function, mirroring the existing query-side pattern. Commands default to requiring authentication (`authenticatedAccess`) — going public is an explicit, single-line opt-in: `canAccess = publicAccess`. The Web transport now falls through to an anonymous context on missing tokens, letting the dispatcher's `canAccessImpl` decide whether that command tolerates unauthenticated callers. Invalid tokens are still rejected immediately, preventing JWT-validity probing.

Closes #638

## Context

Previously, all commands were treated as all-or-nothing at the transport layer: if `Application.withAuth` was wired, every command required a valid JWT regardless of which command. This PR enables:

- **Anonymous feedback form:** A public `SubmitFeedback` command coexists with authenticated commands in the same service.
- **Permission-gated operations:** An `admin:delete` command requires the `admin:delete` permission via `canAccess = requirePermission "admin:delete"`.
- **Public webhooks:** External systems can call framework-managed command endpoints without signing in.

See #638 for the original design discussion.

### Source files modified (5)

1. **`core/service/Service/Command/Core.hs`** — Added `canAccessImpl` method to the `Command` typeclass with default `authenticatedAccess`. Typeclass default enforces whitelist-by-default: a command with no `canAccess` override rejects unauthenticated callers.

2. **`core/service/Service/CommandExecutor/Core.hs`** — Dispatcher calls `canAccessImpl` before `decideImpl`. On authorization failure, returns `CommandUnauthorized` (new `ExecutionResult` variant).

3. **`core/service/Service/CommandExecutor/TH.hs`** — TH macro optionally wires top-level `canAccess` into `canAccessImpl`. When `canAccess` is absent, typeclass default applies (no boilerplate required).

4. **`core/service/Service/Transport/Web.hs`** — Fallthrough on missing token; still rejects invalid tokens. Maps `CommandUnauthorized` to 401/403 responses using the same error matrix as queries.

5. **`core/service/Service/Query/Auth.hs`** — Minor: exports `publicAccess`, `authenticatedAccess`, `requirePermission`, `requireAnyPermission`, `requireAllPermissions` unchanged (they were already accessible via internal imports; now available to commands).

### New files (2)

1. **`core/service/Service/Command/Auth.hs`** — Re-exports `QueryAuthError` as `CommandAuthError` (alias) plus the five permission helpers. No duplication: commands reuse the exact same error type and helper bodies as queries.

2. **`docs/decisions/0058-per-command-auth-options.md`** — ADR with design rationale, considered options, and the complete matrix of request × auth config × response outcomes.

### Tests added (23 new examples across 4 spec files)

- **`core/test/Service/Command/CanAccessSpec.hs`** — TH-resolution tests: commands with no `canAccess` default to `authenticatedAccess`; commands with `canAccess = publicAccess` resolve correctly.
- **`core/test/Service/Transport/Web/CommandAuthSpec.hs`** — Web dispatcher tests: missing token + public → 200 anonymous; invalid token + public → 401; valid token + permission → 200.
- **`core/test/Service/CommandExecutor/AuditLoggingSpec.hs`** — Audit logging: `CommandUnauthorized` cases are logged with command type, claims-present boolean, and error constructor name (no claim contents or token bytes leaked).
- **`core/testlib/Test/Service/CommandHandler/Execute/Spec.hs`** — Updates to existing multitenancy tests to account for the new `canAccessImpl` call site in the dispatcher.

## Backwards compatibility

**Default is `authenticatedAccess`; commands with no `canAccess` declared behave exactly as today.** Existing services are not affected. The Web transport's fallthrough on `TokenMissing` only matters for commands that explicitly opt into public access via `canAccess = publicAccess` or similar. Commands that define nothing inherit the secure default and continue to reject unauthenticated callers.

## Security

Findings from phase 12 (implementation review) were either addressed in the ADR or deferred to test spec (phase 8):

- **Whitelist by default at framework layer:** A junior developer who writes a new command and forgets `canAccess` cannot accidentally ship a public endpoint. The typeclass default forces `authenticatedAccess`.
- **Invalid tokens are never silently accepted:** Malformed, expired, bad-signature, key-not-found, and infra-unavailable errors reject at the middleware before reaching the dispatcher, exactly as on the query side. An attacker cannot use a public command as a JWT-validity oracle.
- **Audit logging on every `CommandUnauthorized`:** Rejection events are logged at `Log.warn` with three fields only: command type name, boolean `claims_present`, and the `CommandAuthError` constructor name. Permission strings, claim contents, token bytes, and JWT subjects are never logged — they would leak sensitive identifiers into the operator's log pipeline.

See `.pipeline/findings-04.json` for the full security review.

## Performance

Findings from phase 13 (performance review) are kept as advisories; none block the PR:

- **Typeclass dispatch overhead on the hot path** (`canAccessImpl @command` once per request): Expected negligible due to GHC specialization at the call site. **Post-merge profiling task:** Verify GHC specialises `canAccessImpl @command` at the dispatcher call site via Core dumps (`-ddump-simpl`) or late-cost-centre profile. Only add `{-# INLINABLE canAccessImpl #-}` if profiling shows >50 ns/call dictionary-passing overhead. Likely a no-op.

- **Case-match branching in the dispatcher:** The new `case canAccessImpl of` has minimal overhead and should be inlined/merged by GHC, but profiling can confirm no instruction-cache pressure.

See `.pipeline/findings-05.json` and `.pipeline/findings-13.json` for details.

## Pre-existing flake warning

**If CI flakes on EventStore ordering tests or `CommandHandler Execute Specification Tests → Concurrency Handling`, re-run.** We observed non-deterministic failure counts (4+2 vs 2+4) on consecutive `cabal test all` runs of unchanged code. Each suite passes cleanly when run serially. The root cause is the shared Postgres instance — separate infra concern, not introduced by this PR. Failure signatures that ARE introduced by this PR would be in `CommandAuthSpec`, `AuditLoggingSpec`, `CanAccessSpec`, or `Service.MultiTenant`.

## Tests

All suites pass with 4184 examples / 0 failures when run serially.

**ADR-0058 coverage:** 23 new examples across:
- `CanAccessSpec` (TH-resolution tests)
- `CommandAuthSpec` (Web dispatcher matrix)
- `AuditLoggingSpec` (logging validation)
- `Service.MultiTenant` (multitenancy integration)

## Hlint

119 hints total, none in new code (all in pre-existing `testbed/` examples).

## Checklist

- [x] ADR approved (phase 3)
- [x] Security findings grounded (phase 12)
- [x] Performance findings grounded (phase 13)
- [x] Tests passing (phase 15)
- Hlint: clean ✓

## Closes

Closes #638

## Release Notes

* **New Features**
  * Added per-command authorization framework with pre-execution access control gates
  * Commands now default to requiring authentication; can be customized to support public, permission-based, or custom access policies
  * Unified authorization error handling across commands and queries with consistent HTTP status codes (401 for unauthenticated, 403 for forbidden)

* **Bug Fixes**
  * Multi-tenant commands now return proper authorization errors instead of generic rejections for missing/invalid tenant context

* **Documentation**
  * Added architectural decision record documenting per-command authorization design and test validation strategy


NickSeagull:
## CodeRabbit review response

Pushed `ecbd22f` addressing the actionable findings. Triage below.

### Fixed in this push

- **`Web.hs:587` — JSON content-type bug.** Wrapped the auth-error body in `{"error": "<msg>"}` so clients parsing `application/json` no longer choke on plain text. (Major / Quick win — agreed and applied.)
- **`CommandExecutor/Core.hs:169,174,179` — tenant auth on the wrong channel.** Multi-tenant tenant-extraction failures now emit `CommandUnauthorized` (→ 401/403) instead of `CommandRejected` (→ 400), matching the per-command auth gate above them. Four expectations in `MultiTenantSpec.hs` updated to track the new behaviour. (Minor consistency fix — agreed and applied.)
- **`TH.hs:501` — stale STUB commentary.** Replaced with the actual contract. (Cosmetic — applied.)
- **`AuditLoggingSpec.hs:88` — PII test was a no-op.** Rewrote the test to actually thread claims (sub, email, name, tenantId, permissions) through `canAccessImpl @CommandWithAdminDelete` and assert the resulting `InsufficientPermissions` payload contains only `["admin:delete"]` — explicitly verifying that sub/email/name/tenantId never appear. (Major — agreed and applied.)

### Deferred — flagged as follow-up, not in scope for this PR

- **`AuditLoggingSpec.hs:77`, `CommandAuthSpec.hs:119`, `CommandAuthSpec.hs:151`** — "must exercise the actual Web transport / capture Log.warn output / verify per-transport drift".

These three are valid quality concerns, but each requires test infrastructure that doesn't exist in this codebase yet:

- Asserting `Log.warn` emission needs a structured-log capture sink wired into the test harness.
- Per-transport tests need a Wai test client, fake JWKS + JWT signer, and (for MCP) an in-process MCP stub harness.

The current tests at the listed lines do verify a real load-bearing property — that `canAccessImpl` resolves correctly via the typeclass for each command policy (default `authenticatedAccess`, explicit `publicAccess`, `requirePermission "..."`). They're shallow on the "actually crosses the transport boundary" axis. Building that infrastructure here would multiply the PR's blast radius beyond the ADR's scope.

Filing as a follow-up. The transport-parity guarantee is structurally enforced today: `canAccessImpl` is called inside the shared dispatcher (`Service.CommandExecutor.Core.execute`), which every transport routes through, so drift can only happen if someone routes around the dispatcher — which is its own architecturally visible change.

### Build state

- `cabal build all` ✅
- PG suites serially: `nhcore-test-service` 827/0 fail, `nhcore-test` 1651/0 fail (parallel-PG flake remains pre-existing, unrelated to this PR — re-run on EventStore-ordering / concurrency hits)

NickSeagull:
## CodeRabbit follow-up: #6 and #7 addressed

Pushed `eac522d`.

### #6 — Web transport coverage

Followed the codebase's existing test pattern (see `Service.Transport.Web.HealthCheckSpec` — it tests `buildHealthResponse` via the pure helper rather than spinning up a Wai application). Applied the same shape here:

- **Extracted** the auth-error mapping from the inline dispatch lambda in `Web.hs` into two pure helpers, `unauthorizedResponse` and `unauthorizedResponseBody`, both exported.
- **CommandAuthSpec.hs Section B** (new) covers them directly:
  - All three `CommandAuthError` constructors → expected `(HTTP.Status, message)` pair.
  - JSON envelope produces `{"error": "..."}` exactly.
  - **Privacy assertion**: `InsufficientPermissions ["admin:delete", "secret:scope"]` → the HTTP body must NOT contain either permission name. The audit log carries that; the wire never does.
  - Defensive: messages with embedded quotes stay JSON-well-formed (envelope still starts with `{"error":` and ends with `}`).

### #7 — Transport parity

The four old cases (each calling `canAccessImpl @CommandWithoutCanAccess Nothing` under different `describe` labels) are gone. Replaced with **Section C** — three tests that call `CommandExecutor.execute` directly with cart fixtures. The argument the comment block makes explicitly:

> The dispatcher is the single entry point all transports route through. Exercising it once proves parity by construction: Web, Cli, Mcp, and Internal each delegate to `CommandExecutor.execute`, so any transport that bypasses the gate would have to bypass this function — an architecturally visible change.

Three cases:
- Anonymous + `authenticatedAccess`-default command → dispatcher emits `CommandUnauthorized Unauthenticated` **before** `decideImpl` runs.
- Anonymous + `publicAccess` command → dispatcher passes through (gate does NOT short-circuit); business reject from the decider proves the open path.
- Authenticated user + `authenticatedAccess`-default command → dispatcher passes through.

### Section A — kept but honestly labelled

The original typeclass-dispatch tests are valuable as TH-wiring verification, so I kept them — but the `describe` block now reads "canAccessImpl resolves correctly per command policy (TH wiring)" instead of pretending to be transport tests. That's the actual scope they cover.

### Build state

- Build green.
- `nhcore-test-service` and `nhcore-test` pass serially with the new tests (28 → +B + +C; 57 in nhcore-test).

core/core/Core.hs
```diff
@@ -31,7 +31,9 @@ import Path as Reexported (Path, path)
 import Record as Reexported (KnownHash (..), Record)
 import Result as Reexported (Result (..))
 import Schema as Reexported (Schema (..), FieldSchema (..), ToSchema (..))
-import Service.Command as Reexported (Command (..), NameOf)
+-- NOTE: canAccessImpl is excluded here to avoid name conflict with Query.canAccessImpl.
+-- Import Service.Command.Core directly to use canAccessImpl on Command.
+import Service.Command as Reexported (Command (IsMultiTenant, getEntityIdImpl, decideImpl), NameOf)
```
NickSeagull: We must then to rename this to something that doesnt clash. All NeoHaskell projects should not require to import anything else that is not `Core`

core/service/Service/Command/Auth.hs
```diff
@@ -0,0 +1,46 @@
...
+  requireAllPermissions,
+  requireAnyPermission,
+  requirePermission,
+ )
+
+
+-- | Alias for 'QueryAuthError' on the command path.
+-- Using a single error type across both query and command authorization
+-- avoids duplication and ensures the same constructors are available everywhere.
+type CommandAuthError = QueryAuthError
```
NickSeagull: I think that rather than doing this level of indirection, we should instead make the AuthError a common type, and then have both QueryAuthError and CommandAuthError users to use that common one, instead of introducing additional names.

core/service/Service/CommandExecutor/Core.hs
```diff
@@ -132,30 +137,52 @@ execute ::
...
+  -- This is a pure check — no event-store calls are made on rejection.
+  case canAccessImpl @command (requestContext.user) of
+    Just authErr -> do
+      -- Audit log: one Log.warn per rejection.
+      -- Fields: command type name, claims_present boolean, constructor name.
+      -- Never log claim contents, permission strings, or token bytes.
+      let commandTypeText = GHC.symbolVal (Record.Proxy @(NameOf command)) |> Text.fromLinkedList
+      let claimsPresent = (case requestContext.user of
+            Nothing -> "false"
+            Just _  -> "true") :: Text
```
NickSeagull: This can be `Maybe.fromDefault`

core/service/Service/Transport/Web.hs
```diff
@@ -556,13 +563,27 @@ instance Transport WebTransport where
...
+                                      |> Bytes.toLazyLegacy
+                                  Maybe.Nothing ->
+                                    responseBytes
+                                      |> Bytes.toLazyLegacy
                             let responseBody =
-                                  responseBytes
-                                    |> Bytes.toLazyLegacy
-                                    |> Wai.responseLBS httpStatus [(HTTP.hContentType, "application/json")]
+                                  finalBodyBytes
+                                    |> Wai.responseLBS statusOverride [(HTTP.hContentType, "application/json")]
```
NickSeagull, line 587: Question here, answer in session: Rather than mirroring, wouldn't it be better to DRY? Or they are not equal 100%?

NickSeagull:
## Addressed @NickSeagull comments — `d6ef981` pushed

### 1. `Core.hs:36` — re-export clash → rename Command method
Renamed the Command typeclass method from `canAccessImpl` to **`canExecuteImpl`**. The user-facing top-level function is still `canAccess` — only the internal method name differs. The new name pairs with the dispatcher entry point (`execute`) and lets `Core` re-export both classes without a conflict.

`Core.hs:36` now reads:
```haskell
import Service.Command as Reexported (Command (IsMultiTenant, getEntityIdImpl, decideImpl, canExecuteImpl), NameOf)
```
NeoHaskell projects no longer need explicit `Service.Command.Core` imports for this method.

### 2. `Service/Command/Auth.hs:46` — common AuthError type
Renamed `QueryAuthError` → **`AccessError`** in `Service.Query.Auth`. Deleted `Service.Command.Auth` (and its cabal entry). Command and query paths now use the same single `AccessError` type directly — no more alias, no more shim.

The module name `Service.Query.Auth` is a bit awkward now that `AccessError` is shared (Command code imports from `Service.Query.Auth`). Worth a follow-up move to e.g. `Service.AccessControl`, but I left that out of this commit so the rename can land cleanly first.

### 3. `CommandExecutor/Core.hs:150` — Maybe helper
Note: there's no `Maybe.fromDefault` in the Maybe module; the closest match is `Maybe.withDefault` paired with `Maybe.map`:
```haskell
let claimsPresent =
      requestContext.user
        |> Maybe.map (\_ -> "true")
        |> Maybe.withDefault ("false" :: Text)
```
If you meant a specific other helper, let me know — happy to swap.

### 4. `Web.hs:586` — DRY between query/command auth fallthrough

You're right that there's real shared shape. Both paths follow:
1. `TokenMissing` → fall through with no-claims context.
2. Any other `AuthError` → reject 401 immediately.
3. `Ok authContext` → pass extracted claims forward.

They differ in **what shape the downstream handler wants**: the query path passes `Maybe UserClaims` directly; the command path needs a built `RequestContext` (`Auth.anonymousContext` / `Auth.authenticatedContext`).

A clean DRY would be a helper that resolves auth into a `Maybe UserClaims` (the LCD), then each call site adapts:
```haskell
-- New helper, used by both query and command paths
withAuthResolved
  :: AuthEnabled
  -> Wai.Request
  -> Respond
  -> (Maybe UserClaims -> Task err Wai.ResponseReceived)
  -> Task err Wai.ResponseReceived

-- Query callsite:
withAuthResolved auth request respond processQueryWithClaims

-- Command callsite:
withAuthResolved auth request respond \claims -> do
  ctx <- case claims of
    Just c  -> Auth.authenticatedContext c
    Nothing -> Auth.anonymousContext
  processCommandWithContext ctx
```

Two ways forward:
- **A**: I implement this DRY refactor as a follow-up commit on this PR.
- **B**: Land this PR with the mirrored shape and file a separate issue for the DRY pass; that keeps the diff scoped to the ADR and lets the refactor land with its own focused review.

Slight preference for **B** — but happy with **A** if you'd rather not split.

### Verification

- Build green.
- `nhcore-test-service`: 830/0 fail.
- `nhcore-test`: 1654/0 fail.

(Net `+3` tests vs the previous push — the cleanup added a couple of cases incidentally.)

NickSeagull:
## #4 (DRY) — implemented as `1871c92`

Extracted `withResolvedClaims` at the bottom of `Web.hs`. Both dispatch sites now share the three-way auth-resolution rule (TokenMissing → fallthrough; other AuthError → reject; valid → pass claims) and only differ in what each does with the resolved `Maybe UserClaims`:

```haskell
-- Command site
withResolvedClaims webTransport.authEnabled request respond \claims -> do
  requestContext <- case claims of
    Maybe.Just c  -> Auth.authenticatedContext c
    Maybe.Nothing -> Auth.anonymousContext
  processCommandWithContext requestContext

-- Query site
withResolvedClaims webTransport.authEnabled request respond processQueryWithClaims
```

The helper is intentionally not exported (Web-transport-internal plumbing). Net `-51 / +49` lines in `Web.hs`.

Build green; PG suites pass serially (830 + 1654 = 2484 tests / 0 failures).

NickSeagull:
## Second-pass CodeRabbit review handled — `dd0ca5e`

### Real bug — `authEnabled = Nothing` regression
**Fixed.** Added `trustedBypass :: Bool` field to `RequestContext` and a `trustedContext` builder. The Web transport now uses `trustedContext` when JWT auth isn't wired (`Application.withAuth` absent). The dispatcher short-circuits the `canExecuteImpl` gate when `trustedBypass = True`, so the legacy no-auth deployment story is preserved: every command stays reachable regardless of its `canAccess` declaration. Only the Web transport's `authEnabled = Nothing` branch constructs this — tests stay on `anonymousContext`/`authenticatedContext`.

Cleaned up `withResolvedClaims` to take `AuthEnabled` (not `Maybe`); the call sites handle the `Nothing` branch explicitly because the two paths actually diverge there (commands use `trustedContext`, queries pass `Nothing` claims as before).

### Test label sed false-positive
`Main.hs:90` relabel: `"Service.Query.Auth"` → `"Service.Command.Auth"`.

### Wildcard pass in dispatcher-parity tests
Tightened the two parity cases to assert the **exact** post-gate outcome (`CommandRejected "Cart does not exist"`). A regression that turns the gate-pass into Accepted, Failed, or a different rejection reason is now caught.

### Style — qualified imports + explicit args
Applied per `CLAUDE.md` "qualified imports + no point-free + no wildcard params" convention to the fixtures and testlib, scoped to files this PR materially touches:
- `PermissionFixture` / `PublicAccessFixture` — `canAccess claims = ...`, `getEntityId _command = Nothing`, `decide _command _entity _context = ...`, `QueryAuth.publicAccess` / `QueryAuth.requirePermission`.
- `testlib/Test/Service/Command/Core.hs` — five `canExecuteImpl claims = QueryAuth.publicAccess claims` bindings.
- `InternalSpec.hs` — same pattern.

Left `AccessError (..)` unqualified in `CommandAuthSpec` / `AuditLoggingSpec` because those files only pattern-match constructors and call no `Service.Query.Auth` functions; the type+constructors import is already the canonical pattern for that use.

### Audit-log header comment
Clarified that `InsufficientPermissions` legitimately carries the permission name on the return value (so an audit emitter can log it). The privacy guarantee is against PII leakage (sub/email/name/tenantId), and a separate test in `Service.Transport.Web.CommandAuthSpec` asserts the HTTP wire body never sees the permission name.

### Out of scope — testbed style debt
The flagged style violations in `testbed/src/Testbed/Cart/Queries/CartSummary.hs` and `Stock/Queries/StockLevel.hs` are pre-existing (this PR only touched them via the `QueryAuthError` → `AccessError` rename pass). Filing as a separate refactor — not in scope here.

### Verification
- Build green.
- `nhcore-test-service`: 830/0 fail.
- `nhcore-test`: 1654/0 fail.

Issue #679, opened 2026-06-22
Epic: Deploy-readiness on Azure Database for PostgreSQL Flexible Server


NickSeagull:
## Summary

We are standardizing the per-client deployment database on **Azure Database for PostgreSQL — Flexible Server** (Burstable `B1ms` baseline) running behind **Azure Container Apps with `minReplicas=0`** (scale-to-zero).

The good news: **connecting nhcore to Flexible Server requires zero code change** — the connection is already fully driven by config (`PostgresEventStore` record at `core/service/Service/EventStore/Postgres/Internal.hs:65-71`, wired from env via `Config.field` in the app layer, e.g. `testbed/src/Testbed/Config.hs:28-45` → `testbed/src/App.hs:64-71`), and no `sslmode` is set anywhere, so libpq's default `prefer` negotiates the TLS that Flexible Server requires.

This epic tracks the **hardening and correctness work in `nhcore`** needed to make that deployment production-grade: surviving Flexible Server's connection budget, maintenance restarts, and failovers without silent data loss — not basic connectivity.

This is a **meta/tracking issue**; each work item below should become (or link to) its own PR. No code is included here by design — only references to the affected code, ADRs, and related issues.

## Sub-issues — suggested execution order

Ordered by dependency and value. Items in the same phase are independent and can be worked in parallel.

**Phase 1 — start here (parallel):**
- [x] #680 — WI-1: Pin `hasql-pool` + explicit pool sizes — connection-budget foundation; deploy-blocking on `B1ms`; unblocks WI-2
- [x] #682 — WI-3: Replay from `globalPosition` on listener reconnect — highest-value correctness (silent data loss); independent

**Phase 2 — after Phase 1 (parallel):**
- [x] #681 — WI-2: Unify connection-settings builders — depends on WI-1 (#680); unblocks WI-5
- [x] #683 — WI-4: Release per-stream connections on unsubscribe — independent correctness; relates to the WI-1 budget

**Phase 3 — after WI-2:**
- [x] #684 — WI-5: Optional `sslmode` TLS hardening — lands in the shared builder from WI-2 (#681)

**Phase 4 — last:**
- [x] #685 — WI-6: Deployment documentation — captures the WI-1 budget + WI-5 hardening outcomes

## Why Flexible Server

[...]

Issue #571, opened 2026-03-16
fix: outbound integration workers lose events on transient network failures (no retry, unhelpful errors)


NickSeagull:
## Summary

Outbound integration workers in the `Dispatcher` crash and silently drop events when a transient network error occurs (e.g., `Connection reset by peer`). There's no retry logic for outbound event processing, and the error message gives no actionable information.

This is a data-loss bug. Events consumed from the channel are gone forever if the worker crashes mid-processing.

---

## Who this affects

**Nick** (integration developer) wires up outbound integrations to emit downstream commands after domain events. When a transient network blip hits, his `ItemAdded -> ReserveStock` chain silently breaks. The app appears stuck. The frontend polls but state never advances. Nick has no idea why.

**Jess** (integration user) sees nothing at all, which is actually the worst outcome: no error, no retry notification, just a frozen UI.

---

### Root cause 1: IO exceptions bypass per-runner error handling

In `processStatelessEvent` (Dispatcher.hs line 694), `Task.asResult` only catches errors raised via `Task.throw`. Raw IO exceptions, like TCP socket errors, propagate straight past it and crash the entire worker through the outer `Task.asResultSafe` boundary.

The same issue exists in `processLifecycleEvent` (line 715).

```
-- Current behavior (simplified)
result <- Task.asResult (runner event)   -- only catches Task.throw
-- ^ raw IOException from network escapes here and kills the whole worker
```

### Root cause 2: No retry for outbound event processing

Even when an error is caught, the event is skipped with no retry. Inbound workers already have exponential backoff retry (`Application/Integrations.hs` lines 319-338). Outbound workers have nothing. This asymmetry means outbound events are far more fragile than inbound ones.

---

## Impact

When a transient network error hits an outbound integration:

1. The event is consumed from the channel but never successfully processed. It's lost.
2. Any downstream command that integration was supposed to emit (e.g., `ReserveStock`) never fires.
3. The app appears stuck with no visible error.
4. The only log output is something like:

```
Stateless worker crashed for stream StreamId "3f2a...": Network.Socket.recvBuf: resource vanished (Connection reset by peer)
```

No entity type. No retry count. No suggestion of what to do next.

---

### 1. Close the exception gap

Replace `Task.asResult` with `Task.asResultSafe` in both `processStatelessEvent` and `processLifecycleEvent`. This ensures raw IO exceptions are caught at the per-runner level instead of crashing the whole worker.

### 2. Add retry with exponential backoff

[...]

Issue #252, opened 2026-01-15
feat: Implement graceful shutdown with drain timeout for dispatcher


NickSeagull:
## Summary

Current shutdown sends `Stop` to workers but doesn't wait for them to finish. In-flight events may be lost.

## User Story

As a developer, I want my application to drain in-flight events during shutdown, so that I don't lose data or cause duplicate processing on restart.

## The Problem

**Current behavior:**
```haskell
shutdown dispatcher = do
  dispatcher.shutdownSignal |> ConcurrentVar.set True
  -- Send Stop to all workers
  -- Returns immediately, workers still running!
```

**Issues:**
1. Stop messages are queued behind pending events
2. Workers may have 100s of events to process
3. Application exits before workers finish
4. Events in queue are lost
5. On restart, events may be reprocessed (duplicates)

## Proposed Solution: Drain with Timeout

```haskell
shutdown :: IntegrationDispatcher -> Int -> Task Text Unit
shutdown dispatcher timeoutMs = do
  -- 1. Stop accepting new events
  dispatcher.shutdownSignal |> ConcurrentVar.set True
  
  -- 2. Send stop to all workers
  allWorkers <- getAllWorkers dispatcher
  allWorkers |> Task.forEach \worker ->
    worker.channel |> Channel.write Stop
  
  -- 3. Wait for workers to finish (with timeout)
  let workerTasks = allWorkers |> Array.map (.workerTask)
  AsyncTask.waitAllWithTimeout workerTasks timeoutMs
  
  -- 4. Force-cancel any remaining workers after timeout
  workerTasks |> Task.forEach \task ->
    AsyncTask.cancel task |> Task.ignoreError
```

**Timeout considerations:**
- Too short: Workers killed mid-processing, events lost
- Too long: Shutdown takes forever, deployment blocked
- **Recommendation**: 30 seconds default, configurable

### Application-Level Coordination

`Application.runWith` should coordinate shutdown:
1. Stop transports (stop accepting new requests)
2. Drain integration dispatcher
3. Stop query subscribers
4. Close event store connections

## Acceptance Criteria

- [ ] `shutdown` function accepts timeout parameter
- [ ] Workers are given time to drain their channels
- [ ] `AsyncTask.waitAllWithTimeout` helper implemented (or equivalent)
- [ ] Workers force-cancelled after timeout expires
- [ ] `Application.runWith` coordinates proper shutdown sequence
- [ ] Unit test: graceful drain completes within timeout
[...]

PR #687, opened 2026-06-24
fix(service): catch-up read on LISTEN/NOTIFY reconnect to prevent missed events

### LINKED ISSUE #682: WI-3: Replay from last globalPosition on LISTEN/NOTIFY listener reconnect
Part of #679.

## What
When the `LISTEN/NOTIFY` listener reconnects, perform a position-based catch-up read from the last processed `globalPosition` before resuming live dispatch, reusing the catch-up path already implemented for subscribe-from-position (the `subscribeToAllEventsFromPosition*` implementation in `core/service/Service/EventStore/Postgres/Internal.hs` — confirm exact entry point during implementation).

## Why
The supervised reconnect loop (`listenerWithReconnect` / `connectTo` in `core/service/Service/EventStore/Postgres/Notifications.hs`) re-issues `LISTEN` on reconnect but performs **no catch-up read**. `NOTIFY` is not durable, so any event inserted while the listener is down (maintenance restart, failover, transient network drop) is **silently skipped from live projections/integrations**. This is a silent correctness bug affecting live dispatch only.

## Acceptance criteria
- A regression test kills the listen connection mid-stream, inserts events during the outage, and asserts **no events are skipped** after reconnect.
- Behavior verified against an induced Flexible Server restart.

## Related
#397 / ADR-0037 (the reconnect loop this builds on), #439 / ADR-0039 (listener lifecycle), #401 (closes one of its silent-failure points), #399 (sibling silent-failure on the command path).

NickSeagull:
## What & why

NeoHaskell's live subscriptions/projections ride on PostgreSQL `LISTEN/NOTIFY`. The supervised reconnect loop re-issues `LISTEN` after a dropped listen connection (maintenance restart, failover, transient network drop) — but `NOTIFY` is not durable, so any event committed **while the listener was down** produced no notification and was **silently skipped** from live dispatch. No error, no log: projections and outbound integrations just missed those events.

## What changed

On every (re)connect, the listener now does a **position-based catch-up read** before resuming live dispatch:

- `Sessions.hs` — new `selectEventsForwardFromGlobalPositionSession` (`WHERE GlobalPosition > $1 ORDER BY GlobalPosition ASC`, exclusive of the cursor).
- `Notifications.hs` — an in-memory `globalPosition` cursor in `connectTo`: initialised "from now" on first connect, advanced by live dispatch, and `catchUpFromCursor` runs **after `LISTEN` is active but before the live wait**. That ordering is load-bearing: the active `LISTEN` and the catch-up read overlap, so an event is either replayed or delivered live — never dropped. Catch-up reuses the existing forward-read path and sinks into the same `SubscriptionStore.dispatch` the live handler uses.
- At-least-once on the reconnect boundary is safe: `dispatch` already filters per-subscription by position and projections are idempotent by `globalPosition`.

### Scope (see ADR-0061)
This covers the **in-process** listener reconnect. A **full service restart** (crash / redeploy / scale-to-zero) loses the in-memory cursor and starts "from now"; read-model projections recover via ADR-0059's persistent QueryObjectStore checkpoints, not this cursor. Outbound-subscriber recovery across a restart is out of scope (tracked by #571 and the scale-to-zero cluster #252/#653/#661/#662).

## Testing

- 11 new `NotificationsSpec` tests, incl. the acceptance test: insert events during a listener outage → assert **no events skipped** after catch-up (plus forward-read, cursor-advance, error-propagation).
- `cabal build all` ✅ · `cabal test all` with PostgreSQL live → **0 failures** across all suites · `hlint` clean.

## References
- ADR: `docs/decisions/0061-listener-reconnect-catchup.md`
- Part of #679 (Azure Flexible Server deploy-readiness epic)

Closes #682

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Postgres `LISTEN/NOTIFY` listener reconnect to replay missed events after outages using the last processed `globalPosition` (exclusive).
  * Ensured the reconnect catch-up cursor advances monotonically for both live notifications and replay, avoiding backward movement.
  * Updated catch-up behavior to surface read/decoding failures instead of silently skipping events.

* **Tests**
  * Added Postgres integration and end-to-end coverage for reconnection/catch-up, including cursor boundaries, ordering, gap replay correctness, and failure propagation.

* **Documentation**
  * Added ADR-0061 documenting the reconnect replay approach and its correctness guarantees.

### DIFF
--- a/core/service/Service/EventStore/Postgres/Notifications.hs
+++ b/core/service/Service/EventStore/Postgres/Notifications.hs
@@ -5,6 +5,7 @@ module Service.EventStore.Postgres.Notifications (
   catchUpFromCursor,
 ) where
 
+import Array qualified
 import AsyncTask qualified
 import Bytes qualified
 import Core hiding (Var)
@@ -14,8 +15,8 @@ import Hasql.Notifications qualified as HasqlNotifications
 import Json qualified
 import Log qualified
 import Maybe (Maybe (..))
-import Service.Event (Event (..), StreamPosition (..))
-import Service.EventStore.Postgres.PostgresEventRecord (PostgresEventRecord (..))
+import Service.Event (Event (..), StreamPosition (StreamPosition))
+import Service.EventStore.Postgres.PostgresEventRecord (PostgresEventRecord (globalPosition))
 import Service.EventStore.Postgres.Sessions qualified as Sessions
 import Service.EventStore.Postgres.SubscriptionStore (SubscriptionStore)
 import Service.EventStore.Postgres.SubscriptionStore qualified as SubscriptionStore
@@ -199,12 +200,28 @@ initialiseOrCatchUp queryConnection store lastProcessedRef initialisedRef = do
       catchUpFromCursor queryConnection store lastProcessedRef
 
 
+-- | Catch-up batch size: events read per forward page during a reconnect
+-- catch-up. Matches the ADR-0059 query-rebuild chunk size (1000) so the two
+-- forward-replay paths page consistently and the catch-up read stays bounded
+-- regardless of how long the listener was offline.
+catchUpBatchSize :: Int64
+catchUpBatchSize = 1000
+
+
 -- | Replay every event committed after the last-processed @globalPosition@
 -- through the shared 'SubscriptionStore', advancing the cursor as it goes.
 -- Invoked on each listener reconnect, AFTER @LISTEN@ is re-established and
 -- BEFORE the live wait, so the gap created while the listen socket was down
 -- reaches live dispatch (ADR-0061). The dispatch sink is identical to the
 -- live handler's, so there is one sink and one read implementation.
+--
+-- The read is BOUNDED: rather than pulling every row after the cursor into one
+-- 'Array' (which a long outage could blow up), it reads in pages of
+-- 'catchUpBatchSize', dispatches each page, advances the cursor to the last
+-- 'globalPosition' in the page (via 'dispatchCaughtUpRecord'), and repeats until
+-- a page returns fewer rows than the limit (the gap is drained). Paging is
+-- monotonic in @globalPosition@, so at-least-once/idempotency are preserved and
+-- ordering is unchanged.
 catchUpFromCursor ::
   Hasql.Connection ->
   SubscriptionStore ->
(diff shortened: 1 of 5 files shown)


NickSeagull:
## CodeRabbit review fixes — commit `f42d527`

- ✅ **Explicit/qualified imports** in `Notifications.hs`.
- ✅ **Bounded the forward catch-up read** — `selectEventsForwardFromGlobalPositionSession` now takes a `LIMIT`, and `catchUpFromCursor` loops in **1000-row batches** (matching the ADR-0059 rebuild chunk), advancing the cursor per page. A reconnect after a long outage no longer reads everything into memory at once.
- ✅ **True reconnect-path regression (D1)** — drives the real `connectTo` / `listenerWithReconnect` loop: captures the listener's backend PID, forces a genuine live drop via `pg_terminate_backend`, inserts events during the outage, and asserts all missed events are caught up after reconnect (verified 5/5 stable runs). Not a wiring-contract stand-in.
- ✅ **Test-helper connection cleanup** — `catchUpFixture` and `insertN` now bracket their connection/store with `Task.finally`, so a setup failure can't leak and poison other Postgres specs.
- ✅ **ADR-0061 aligned** to the shipped `selectEventsForwardFromGlobalPositionSession` in `Sessions.hs` (was incorrectly citing `readAllEventsForwardFrom`/`Internal.hs`).

All 7 suites green (0 failures, Postgres live), `hlint` clean.

NickSeagull:
## Round 2 — reconnect-test hardening (commit `2bf3ea2`)

Addressed the re-review's refinements to the D1 reconnect regression:

- **Deterministic outage window** — a sentinel "listener-ready" event (block until dispatched, proving `LISTEN` is up + cursor advanced) plus a **reconnect gate** (`ConcurrentVar`) that holds the second connect from re-`LISTEN`-ing until after the outage inserts. This guarantees the outage events arrive via **catch-up**, not live delivery — so the test actually proves the fix.
- **Assert the specific outage event IDs** (dispatched ⊇ outage ids), not just `count == 3`. This surfaced a real subtlety: the dispatched `eventId` comes from the stored metadata JSON, not the insertion id — precisely the gap the count-based assertion hid.
- **Honor `pg_terminate_backend`** — throw if it didn't actually terminate the backend.
- Tightened the new imports.

**D1 stability: 8/8** deterministic runs. `cabal test all` (Postgres live): **0 failures** across suites. `hlint` clean.

Issue #685, opened 2026-06-22
WI-6: Deployment docs for Azure Database for PostgreSQL Flexible Server


NickSeagull:
Part of #679.

## What
Document the Flexible Server golden path: env-only connection config, the **direct-`5432` rule** (never the built-in PgBouncer pooler on `6432`, which silently breaks `LISTEN/NOTIFY`), the `B1ms` connection budget and pool-size guidance (#679 WI-1), the ACA scale-to-zero pairing (app scales to zero, DB stays on; the DB does not auto-wake on a request), storage-growth cost behavior, and optional TLS hardening (#679 WI-5).

## Why
This is the operational knowledge that turns the hardening work into a repeatable per-client deployment.

## Acceptance criteria
- Covered in the deployment and Postgres-event-store guides.

## Related / home
#535 (`guides/deployment.mdx`), #536 (`guides/postgresql-event-store.mdx`), #36 (general Deployment features).

PR #691, opened 2026-06-24
fix(service): release per-stream subscription connection on unsubscribe

### LINKED ISSUE #683: WI-4: Release per-stream subscription connections on unsubscribe
Part of #679.

## What
Ensure the dedicated unpooled connection opened per stream subscription is **released on unsubscribe**, and document/cap the number of concurrent stream subscriptions against the tier connection budget.

## Why
`subscribeToStreamEventsImpl` (`core/service/Service/EventStore/Postgres/Internal.hs:771-790`) acquires a raw connection per stream (`:775`), but `unsubscribeImpl` (`:793-797`) only removes the subscription-store entry — it does **not release that connection**. Under stream-subscription churn this accumulates connections against the `B1ms` ~35-usable ceiling.

## Acceptance criteria
- Unsubscribe releases the associated connection; no leak across subscribe/unsubscribe cycles (verified by test).
- Guidance documented on max concurrent stream subscriptions per tier.

## Related
#439 / ADR-0039 (the `close`/cleanup lifecycle precedent this mirrors). Coordinates with #679 WI-1 (connection budget).

NickSeagull:
## Summary

When you subscribe to a single stream with `subscribeToStreamEvents`, the Postgres
event store opens a **dedicated database connection** just for that stream's
`LISTEN`/`NOTIFY` channel. Until now, `unsubscribe` removed the in-memory
subscription but **never closed that connection** — so every
subscribe/unsubscribe cycle leaked one Postgres connection that stayed open until
the whole process exited.

On a small deploy target (Azure Database for PostgreSQL Flexible Server `B1ms`,
~35 usable connections) that leak is a hard outage: a few dozen short-lived
stream subscriptions exhaust the connection ceiling and Postgres starts rejecting
new connections.

This PR ties each stream subscription's dedicated connection to the
`SubscriptionId` you already hold, and releases it automatically when you
unsubscribe. **Nothing changes in how you call the API** — you subscribe and
unsubscribe exactly as before; the cleanup happens for you.

Classification: **moderate** — modifies subscription lifecycle in the existing
Postgres event store (external IO, no new dependency, no auth/secret handling, no
schema change).

## ADR

[ADR-0063 — Release per-stream subscription connections on unsubscribe](docs/decisions/0063-per-stream-subscription-connection-release.md)
(approved). Chosen option: store a per-subscription cleanup `Task` in the
`SubscriptionStore` and run it best-effort on remove — mirroring ADR-0039's
connection-cleanup lifecycle one level down.

## Changes

- **`SubscriptionStore.hs`**
  - `SubscriptionInfo` gains an `onRemove :: Task Text Unit` cleanup field
    (a no-op for global/entity/plain-stream subscriptions).
  - Dropped the derived `Show` and added a hand-written `Show` that elides the
    `Task`/closure fields (a `Task` has no `Show`).
  - New `addStreamSubscriptionWithCleanup` registers a stream subscription
    **together with** its connection-release action, so a connection can never be
    tracked without its cleanup.
  - `removeSubscription` now reads the subscription's cleanup before deleting the
    entry, then runs it best-effort under `Task.recover` + `Log.warn`, so
    `unsubscribe` stays total even if the connection was already dead.
- **`Internal.hs`** — `subscribeToStreamEventsImpl` builds
  `Hasql.release connection` and registers it as the subscription's cleanup.
  `unsubscribeImpl` is unchanged — it already routes through `removeSubscription`,
  which is now the single release funnel.

No public API changed. The new surface (`onRemove`,
`addStreamSubscriptionWithCleanup`) is internal to the Postgres backend.

## Security

ADR-grounded (`findings-04`) and implementation-grounded (`findings-12`) reviews:
**0 blockers**. The new hand-written `Show` is security-positive — it never
renders the connection handle or callback. The `findOnRemove`/`removeSubscription`
sequence is internal, keyed by server-minted `SubscriptionId`s (not attacker
input), and the cleanup is *taken* (read-then-delete) so a double-remove can't
release a connection twice.

## Performance

ADR-grounded (`findings-05`) and implementation-grounded (`findings-13`) reviews:
**0 blockers**, target 50k req/s. No new per-event work — the dispatch hot path
never reads `onRemove`. The cleanup is one `Hasql.release` per unsubscribe (work
the runtime was already obligated to do). `findOnRemove`'s lookup runs only at
unsubscribe and is bounded by the documented concurrency budget.

Advisory capacity note (from ADR-0060's `19 + N ≤ 35` budget): keep concurrent
stream subscriptions at or below **~14 on `B1ms`**. This is documented guidance,
**not** a hard runtime limit — nhcore does not reject the 15th subscription; the
number is carried into the deploy guide (WI-6, #685) and a tier upgrade raises it.

## Tests

- **Unit (`SubscriptionStoreSpec.hs`, no Postgres):** cleanup runs exactly once on
  remove; `unsubscribe` stays total when the cleanup throws; removing an unknown
  id is a no-op that runs no cleanup; a double-remove does not re-run the cleanup;
  global/entity subscriptions' no-op cleanup is unaffected; the new constructor
  registers and dispatches like any other stream subscription and does not fire
  cleanup at registration time.
- **Regression (`PostgresSpec.hs`, Postgres-gated on `POSTGRES_AVAILABLE`):** after
  a warm-up, runs N stream subscribe→unsubscribe cycles and asserts the backend
  connection count (`pg_stat_activity`) does not grow — proving the dedicated
  connection is released each cycle and the leak cannot silently return.

All suites green: **4415 examples, 0 failures** (lsp, auth, integrations,
core, service, and the full `nhcore-test`).

## Hlint

Hlint clean.

## Checklist

- [x] ADR approved
- [x] Security findings grounded (0 blockers)
- [x] Performance findings grounded (0 blockers)
- [x] Tests passing (build + full suite green)
- Hlint: clean ✓

Closes #683

## Summary by CodeRabbit

* **New Features**
  * Stream subscriptions now support per-subscription cleanup logic tied to unsubscription.
* **Bug Fixes**
  * Fixed a PostgreSQL connection leak during repeated subscribe/unsubscribe cycles.
  * Subscription removal is now safer: cleanup runs exactly once (best-effort), failures don’t break removal, and repeated/unknown removals are no-ops.
* **Tests**
  * Added unit and PostgreSQL-gated regression coverage for cleanup semantics and connection release across churn.
* **Documentation**
  * Added an ADR describing the per-stream connection release fix.

### DIFF
--- a/core/service/Service/EventStore/Postgres/SubscriptionStore.hs
+++ b/core/service/Service/EventStore/Postgres/SubscriptionStore.hs
@@ -6,6 +6,7 @@ module Service.EventStore.Postgres.SubscriptionStore (
   addGlobalSubscriptionFromPosition,
   addStreamSubscription,
   addStreamSubscriptionFromPosition,
+  addStreamSubscriptionWithCleanup,
   addEntitySubscription,
   addEntitySubscriptionFromPosition,
   getStreamSubscriptions,
@@ -21,6 +22,7 @@ import Core
 import Json qualified
 import Log qualified
 import Map qualified
+import Prelude qualified
 import Service.Event (EntityName, Event (..), StreamPosition)
 import Service.Event.EventMetadata (EventMetadata (..))
 import Service.EventStore.Core (SubscriptionId (..))
@@ -41,9 +43,23 @@ type SubscriptionCallback =
 data SubscriptionInfo = SubscriptionInfo
   { callback :: SubscriptionCallback,
     startingGlobalPosition :: Maybe StreamPosition,
-    entityNameFilter :: Maybe EntityName
+    entityNameFilter :: Maybe EntityName,
+    -- | Run once when this subscription is removed. Releases any dedicated
+    -- resource the subscription owns (the per-stream LISTEN connection). A
+    -- no-op ('Task.yield unit') for subscriptions that own no dedicated resource.
+    onRemove :: Task Text Unit
   }
-  deriving (Show)
+
+
+-- Hand-written Show: the callback and onRemove are Task/closure fields with no
+-- Show instance, so they are elided. (ADR-0063 §1.)
+instance Show SubscriptionInfo where
+  show info =
+    "SubscriptionInfo {startingGlobalPosition = "
+      ++ Prelude.show info.startingGlobalPosition
+      ++ ", entityNameFilter = "
+      ++ Prelude.show info.entityNameFilter
+      ++ ", callback = <function>, onRemove = <task>}"
 
 
 type Subscriptions =
@@ -68,7 +84,7 @@ new = do
 addGlobalSubscription :: SubscriptionCallback -> SubscriptionStore -> Task Error SubscriptionId
 addGlobalSubscription callback store = do
   subId <- Uuid.generate |> Task.map (\result -> result |> toText |> SubscriptionId)
-  let subscriptionInfo = SubscriptionInfo {callback, startingGlobalPosition = Nothing, entityNameFilter = Nothing}
+  let subscriptionInfo = SubscriptionInfo {callback, startingGlobalPosition = Nothing, entityNameFilter = Nothing, onRemove = Task.yield unit}
   store.globalSubscriptions
     |> ConcurrentVar.modify (Map.set subId subscriptionInfo)
   Task.yield subId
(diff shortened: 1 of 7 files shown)


core/service/Service/EventStore/Postgres/SubscriptionStore.hs
```diff
@@ -41,9 +43,23 @@ type SubscriptionCallback =
...
+
+-- Hand-written Show: the callback and onRemove are Task/closure fields with no
+-- Show instance, so they are elided. (ADR-0063 §1.)
+instance Show SubscriptionInfo where
+  show info =
+    "SubscriptionInfo {startingGlobalPosition = "
+      ++ Prelude.show info.startingGlobalPosition
+      ++ ", entityNameFilter = "
+      ++ Prelude.show info.entityNameFilter
+      ++ ", callback = <function>, onRemove = <task>}"
```
NickSeagull: Use `toText` not `show`. Also, use `[fmt|...|]` directly.

Even further, I'd rely directly on the `deriving Show` behavior and would add the Show instance for Task, do this instead

Issue #684, opened 2026-06-22
WI-5: Optional sslmode TLS hardening for Postgres connections


NickSeagull:
Part of #679.

## What
Provide an **optional** way to set `sslmode=require` (and ideally `verify-full` with an `sslrootcert` bundling DigiCert Global Root G2 + Microsoft RSA Root CA 2017, trusting roots only), threaded through the shared builder from #679 WI-2, gated behind config so localhost/dev without TLS still works.

## Why
Not required to connect — libpq's default `prefer` already uses TLS against Flexible Server. But `require`/`verify-full` forbids silent downgrade and is good production hygiene. **Must not** add client certs: Azure does not support mTLS and `sslcert`/`sslkey` will break the connection.

## Acceptance criteria
- Hardening is opt-in via config; defaults preserve current dev behavior; documented.
- Root-only trust (no intermediate/server pinning, given Microsoft's ongoing intermediate-CA rotation).

## Related
Depends on #679 WI-2 (lands in the shared builder). #644 is a separate outbound-HTTP TLS concern — not this path.

PR #695, opened 2026-06-26
fix(service): non-unset sslMode no longer drops the Postgres host param (#694)

### LINKED ISSUE #694: Postgres: non-unset sslMode drops the host param (connection falls back to local unix socket)
## Summary

Setting a non-`unset` Postgres TLS mode (`DB_SSL_MODE=require`, or any `SslMode` other than `SslModeUnset` on `PostgresEventStore`) makes the connection **drop the `host` parameter** and fall back to a local Unix socket. The app then fails to start with:

```
TableInitializationError "ConnectionUsageError (Just \"connection to server on socket \\\"/run/postgresql/.s.PGSQL.5432\\\" failed: No such file or directory\nIs the server running locally and accepting connections on that socket?\n\")"
```

The same configuration connects correctly when `DB_SSL_MODE` is left `unset`. This makes the documented TLS-hardening path (`require` / `verify-ca` / `verify-full`, ADR-0064 / #684) unusable, including the "golden path" `DB_SSL_MODE=require` from the Azure Database for PostgreSQL Flexible Server guide.

## Root cause

In `core/service/Service/Infra/Postgres/ConnectionConfig.hs`, `toConnectionParams` returns a **list of `Setting`s**, where the base params and each SSL param are emitted as *separate* `ConnectionSetting.connection` values:

```haskell
toConnectionParams cfg =
  cfg |> resolveParams |> Result.map \resolved -> do
    let baseParams = [ Param.host resolved.host, Param.port resolved.port, ... ]
    let baseSetting = ConnectionSettingConnection.params baseParams |> ConnectionSetting.connection
    baseSetting : sslParams resolved          -- <-- two `connection` settings

sslParams resolved =
  case sslModeToText resolved.sslMode of
    Nothing    -> []
    Just token -> paramSetting (Param.other "sslmode" token) : rootCertParams ...

paramSetting p =
  ConnectionSettingConnection.params [p] |> ConnectionSetting.connection   -- <-- host-less `connection`
```

`Hasql.Pool.Config.staticConnectionSettings` is then given `[baseSetting, sslmodeSetting]`. With multiple `Hasql.Connection.Setting.connection` entries, the **last one wins** rather than being merged — and the `sslmode` setting contains *only* `sslmode` (no `host`/`port`/`dbname`/`user`/`password`). So the host is lost and libpq falls back to its compiled-in default Unix socket.

When `sslMode == SslModeUnset`, `sslParams` returns `[]`, only `baseSetting` is present, and the host is honored — which is why `unset` works and any set mode breaks.

## Reproduction

1. Configure a `PostgresEventStore` with `sslMode = SslModeRequire` (e.g. `DB_SSL_MODE=require`) and a **remote** `host` (`DB_HOST=db.example.com`).
2. Start the app.
3. Observe the Unix-socket connection error above — the remote host is never contacted.
4. Set `DB_SSL_MODE=unset` (or omit it) with the same host → libpq resolves/contacts the host (e.g. `could not translate host name "db.example.com" ...`), confirming the host is only honored without an SSL mode.

(Reproduced by running the native binary directly with `env -i DB_HOST=... DB_SSL_MODE=require ./app`.)

## Expected

All libpq params — base **and** the conditional `sslmode` / `sslrootcert` / `channel_binding` — should be combined into a **single** `ConnectionSettingConnection.params [...]` (one `ConnectionSetting.connection`), so the SSL params extend the base set instead of replacing it.

## Suggested fix

Concatenate the param lists before building the single connection setting, e.g.:

```haskell
toConnectionParams cfg =
  cfg |> resolveParams |> Result.map \resolved ->
    [ ConnectionSettingConnection.params (baseParams resolved ++ sslParamList resolved)
        |> ConnectionSetting.connection
    ]
```

where `sslParamList` returns the bare `[Param]` (`sslmode`, and for verifying modes `sslrootcert` / `channel_binding`) rather than wrapping each in its own `connection` setting.

## Affected version

Reproduced at `97cef9a8e56a144a84ca94822dacd35011c199a1` (introduced by #684 / ADR-0064 "optional sslmode TLS hardening"). Also affects the FileUpload and QueryObjectStore pools, which route through the same builder.

NickSeagull:
## Summary

Fixes #694. Setting a non-`unset` Postgres TLS mode (`DB_SSL_MODE=require`, or any `SslMode` other than `SslModeUnset`) silently **dropped the `host`/`port`/`dbname` params** and libpq fell back to the local Unix socket, so the app failed to start against a remote host:

```
TableInitializationError "ConnectionUsageError (Just \"connection to server on socket \\\"/run/postgresql/.s.PGSQL.5432\\\" failed: No such file or directory ...\")"
```

This made the documented TLS-hardening path (ADR-0064 / #684) — including the `DB_SSL_MODE=require` golden path from the Azure Database for PostgreSQL Flexible Server guide (#685) — unusable. Only `unset` worked.

## Root cause

`Service.Infra.Postgres.ConnectionConfig.toConnectionParams` emitted the base params and **each** TLS param as *separate* `Hasql.Connection.Setting.connection` values. hasql's `Hasql.Pool.Config.staticConnectionSettings` applies each `connection` Setting by **replacing** the whole connection string rather than merging:

```haskell
-- hasql 1.9.3.2, Hasql.Connection.Config
setConnectionString s config = config {connectionString = s}   -- replace
fromUpdates = foldl' (flip update) nil                          -- last wins
```

So `[baseSetting, sslmodeSetting]` rendered down to just `sslmode='require'` — host/port/dbname gone. `SslModeUnset` emitted no TLS setting, which is exactly why the unset path was the only one that worked.

## Fix

Build every libpq param — base **and** the conditional `sslmode` / `sslrootcert` / `channel_binding` — into **one** `ConnectionSettingConnection.params [...]`, i.e. a single `connection` Setting. All three pools (EventStore, FileUpload, QueryObjectStore) route through this single builder, so all three are fixed at once.

- The `unset` path stays **byte-identical** (the same 9 params in one setting), preserving the ADR-0064 §2 no-regression guarantee.
- `Param.other k v` is byte-identical to the typed `Param.host`/`Param.port`/… constructors (both call `setKeyValue`), so the rendered string is unchanged apart from the now-included TLS params.

## Reproducing the bug in tests

The hasql `Param`/`Setting` types have no `Eq`/`Show` and the modules that render them to a connection string are **internal** to hasql, so a test cannot inspect the final string. The pre-existing tests used `LinkedList.length` of the settings list as a proxy — which is precisely what masked the bug (more TLS params ⇒ more *separate* settings looked "correct").

This PR adds an inspectable `toParamPairs :: ConnectionParams -> Result Text (LinkedList (Text, Text))` — the canonical, ordered `(key, value)` param set the single Setting is built from. Tests now assert the contract **directly**:

- **#694 regression** (`toConnectionParams`): emits exactly **one** connection Setting for *every* `SslMode` (a 2nd `connection` Setting is the host-dropping smoking gun).
- **host survival** (`toParamPairs`): the `host` param is present for `require`, `verify-full`, and every mode (table sweep).
- **exact param sets**: `verify-full + cert` emits exactly `sslmode` + `sslrootcert` + `channel_binding` and **no** `sslcert`/`sslkey` (ADR-0064 §3 no-mTLS guarantee), root-cert is gated to verifying modes only, etc.

**Red verified:** temporarily reintroducing the split-setting structure makes `toConnectionParams` return 2 settings for `require` and 4 for `verify-full`+cert, and the regression tests fail exactly as designed. Restoring the fix → all green (`ConnectionConfigSpec` 55/55; FileUpload sslMode-threading 5/5).

> A true end-to-end TLS connection test needs a remote TLS Postgres (not available in CI). The bug lives entirely in connection-string construction, so these construction-level unit tests catch it deterministically and fast.

## ⚠️ Changed test expectations (needs maintainer sign-off)

Per CONTRIBUTING/CLAUDE rules I'm flagging that two existing expectations changed, because they **encoded the buggy multi-setting shape** introduced by the same #684 PR:

- `ConnectionConfigSpec`: the length-based ssl tests (`baseLen + 1/2/3`) are replaced by direct `toParamPairs` key/value assertions (strictly more precise).
- `FileUpload/FileStateStore/PostgresSpec`: `... |> toConnectionParams |> Result.map LinkedList.length |> shouldBe (Ok 2)` → `(Ok 1)`, plus a `toParamPairs` check that `sslmode=require` is actually threaded through.

## Verification

- `cabal test nhcore-test-service --test-options='--match "Service.Infra.Postgres.ConnectionConfig"'` → 55 examples, 0 failures
- `cabal test nhcore-test --test-options='--match "sslMode threading"'` → 5 examples, 0 failures
- `nhcore-test` full suite compiles (all callers + 99 test modules link)
- `hlint` → No hints on all changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Postgres connection settings so enabling TLS no longer overrides or drops essential base connection details.
  * Improved TLS/SSL parameter generation for `require`, `verify-ca`, and `verify-full`, including correct inclusion/exclusion of root certificate and channel binding settings depending on the mode.

* **Tests**
  * Expanded regression coverage for Postgres TLS parameter generation and ensured the produced parameters match expected behavior across all `SslMode` variants.

### DIFF
--- a/core/service/Service/Infra/Postgres/ConnectionConfig.hs
+++ b/core/service/Service/Infra/Postgres/ConnectionConfig.hs
@@ -6,6 +6,7 @@ module Service.Infra.Postgres.ConnectionConfig (
   sslModeToText,
   textToSslMode,
   resolveParams,
+  toParamPairs,
   toConnectionParams,
   toPoolConfig,
   logPoolObservation,
@@ -22,6 +23,7 @@ import Hasql.Connection.Setting.Connection.Param qualified as Param
 import Hasql.Pool.Config (Config)
 import Hasql.Pool.Config qualified as HasqlPoolConfig
 import Hasql.Pool.Observation (ConnectionStatus (..), ConnectionTerminationReason (..), Observation (..))
+import LinkedList qualified
 import Log qualified
 import Result qualified
 import Task qualified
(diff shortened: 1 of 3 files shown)


NickSeagull:
Thanks @coderabbitai — both findings addressed in 57b3c70:

1. **ConnectionConfigSpec base-key sweep** now covers the hardened CA-bundle path: the table sweep includes `(verify-ca, Just "/etc/ca.pem")` and `(verify-full, Just "/etc/ca.pem")` alongside the no-CA cases, so it proves `host`/`port`/`dbname`/`user`/`password` survive even once `sslrootcert` (+ `channel_binding`) are appended.
2. **FileUpload FileStateStore PostgresSpec** adds a `verify-full + sslrootcert + channel_binding` threading check, so `toEventStoreConfig` can't silently drop the CA bundle while the sslmode-only assertion stays green.

Verified green: `ConnectionConfigSpec` 55/55, FileUpload sslMode-threading 6/6, hlint clean.

@coderabbitai review

PR #708, opened 2026-07-02
feat(web): serve static SPA assets from WebTransport (#707)

### LINKED ISSUE #707: WebTransport: serve static assets (SPA) with index.html fallback
## Summary

Add a way for `WebTransport` to serve **static assets** (a directory of files, or embedded bundles) with **SPA `index.html` fallback**, so an application can serve a single-page frontend from the same transport that already serves its command/query API — without standing up a second WAI/Warp server.

## Motivation

Tooling built on NeoHaskell increasingly wants to ship a web UI alongside its HTTP API (concretely: an event-modeling IDE whose React SPA talks to the app's commands/queries). Today that requires running a separate static server next to `WebTransport`, duplicating port/config/lifecycle management, when the transport is *already* a Warp server that could serve the bundle directly.

The command/query surface already maps cleanly onto WebTransport:

- `GET /queries/<name>` for reads
- `POST /commands/<name>` for writes

The only missing piece to host a full app (API + UI) on one transport is static-asset serving.

## Current behavior

`Service.Transport.Web` dispatches on a **closed** route set:

```haskell
-- core/service/Service/Transport/Web.hs:555
case Wai.pathInfo request of
  ["commands", commandNameKebab] -> ...
  ["queries",  queryNameKebab]   -> ...
  -- health / ready / oauth2 / files / swagger-ui ...
  _ -> notFound "Not found"   -- Web.hs:1026
```

There is no static-file branch and no user-extensible raw-route hook — an unmatched path always 404s.

Notably, the framework already has a precedent for serving static content from a WebTransport branch: the Swagger/Scalar UI is served as `text/html` from a dedicated branch (`SwaggerUI.scalarHtml`, ~Web.hs:1000). This proposal generalizes that pattern into first-class static-asset serving.

## Proposal

Add opt-in static serving, configured through the `Application` builder, e.g.:

```haskell
Application.new
  |> Application.withStaticAssets StaticAssets
       { root      = "dist/"          -- directory on disk, or embedded via file-embed
       , spaFallback = Just "index.html"
       }
  |> ...
```

Behavior:

1. **Precedence** — static serving is the *fallback*, evaluated only after the existing API/health/oauth2/files/swagger branches, i.e. it replaces the terminal `notFound "Not found"` with "try to serve a static file, else 404". API routes always win over files.
2. **SPA fallback** — when `spaFallback` is set and the requested path doesn't resolve to a real file (and isn't an API route), serve the fallback document (typically `index.html`) so client-side routing works on deep links.
3. **Cache-Control** (load-bearing — please don't omit): hashed, content-addressed bundles (`/assets/*.[hash].js` etc.) should be served `immutable`, while `index.html` and other non-hashed entry documents must be `no-cache, must-revalidate`. Getting this wrong makes browsers show a stale UI after every rebuild; getting it right is what makes hot-rebuild-during-dev usable.
4. **No new dependencies required** — WAI can serve files via `responseFile`, and Warp streams response bodies natively, so this need not pull in `wai-app-static` (though that remains an option if preferred).
5. **Path traversal safety** — normalize/'..'-guard requested paths against `root`.

## Out of scope

- Server→client push (SSE / websockets) — tracked separately.
- Authn/authz for static assets (initial version can serve them unauthenticated; revisit if a use case needs gating).

---

NickSeagull:
## Summary

This PR adds `Application.withStaticAssets`, a web-transport combinator that serves static Single-Page Application (SPA) assets from disk. A developer builds the config record `StaticAssets { root = "static", spaFallback = Just "index.html" }` and pipes it once at Application setup — the framework then serves files from `static/` on every unmatched GET, with index.html as the deep-link fallback for SPA client-side routing. Cache-Control headers are set correctly out of the box (content-hashed bundles → immutable, index.html → no-cache), path traversal is guarded with RFC-3986 normalization, and the feature is opt-in (missing/empty root → loud warning, API keeps serving). No new dependencies; reuses WAI's `responseFile` for streaming.

**Tier:** Moderate (filesystem IO on the static-fallback branch, no new dependencies, all path reads validated/normalized)

## Decision & Changes

See [ADR-0066](https://github.com/neohaskell/NeoHaskell/blob/main/docs/decisions/0066-static-assets.md) for design, use cases, and options rejected.

**Public API** (`Service.Transport.Web` — Jess's surface):
- `StaticAssets { root :: Text, spaFallback :: Maybe Text }` — config record
- `Application.withStaticAssets @config (config -> StaticAssets) -> Application -> Application` — one-line combinator, same shape as `withCors`, `withApiInfo`

**Behaviour**:
- Missing or empty `root` → loud startup `Log.warn`, API continues serving normally
- Unmatched request path → terminal fallback: if `spaFallback = Just "index.html"`, serve that for deep links; otherwise 404
- All `..` segments in the request path or fallback path are rejected (404, no filesystem confirmation)
- Path normalization per RFC-3986 before joining to root
- Cache-Control: `immutable` for files matching `name.<hash>.ext` pattern (content-hashed bundles), `no-cache, must-revalidate` for everything else (index.html, entry docs)
- Serves via `responseFile` (zero-copy sendfile on supported systems)

**Internal** (`Service.Transport.Web` — not separately exported):
- `serveStaticAsset` — handles a single request, checks traversal guards, sets Cache-Control, calls `responseFile`
- `serveStaticSpaFallback` — as above but for the fallback doc (same normalization + prefix guard as request path)

## Security

[Phase 12 security findings](https://github.com/neohaskell/NeoHaskell/pull/708) identified one critical issue during implementation review; **fixed**:

**SEC-010 — SPA fallback path traversal:** The `spaFallback` parameter (developer-supplied config) was not validated against directory escape like request paths. Implemented: the fallback path undergoes identical normalization and root-prefix verification as request paths. Regression tests 4.6 and 4.7 added and passing; both validate escape attempts.

Informational findings (not blocking):
- **SEC-001 (kept):** Path traversal on request paths is framework-absorbed inside `withStaticAssets`; regression tests verify both request and fallback paths (4.6/4.7)
- **SEC-011 (informational):** File existence check per request (TOCTOU window) is acceptable; documented risk on case-sensitive filesystems (Linux, deploy target)
- **SEC-012–SEC-014 (informational):** Path-traversal guard verified; symlink risk is deployment-level; dotfiles not served in v1

## Performance

[Phase 13 performance findings](https://github.com/neohaskell/NeoHaskell/pull/708) reviewed 18 patterns; all grounded as informational or framework-debt follow-ups:

- **PERF-011 (advisory):** Per-request `listDirectory` on case-sensitive filesystems is pure overhead (deploy target is Linux where `doesFileExist` is already case-sensitive). Documented follow-up: gate behind macOS-only startup flag; not a blocker on v1.
- Strict-field pragmas, INLINE recommendations, and encoder patterns are either already applied correctly or are deferred to framework-level optimizations pending profiling evidence.

## Tests

Comprehensive test coverage with **43 StaticAssets examples**:
- **4.1–4.5:** Happy path (serve file, deep link, fallback miss), cache-control correctness (hashed vs non-hashed), edge cases (empty root, missing fallback)
- **4.6–4.7:** Security regressions — path traversal on both request and fallback paths (escape attempt → 404, no confirmation)
- **4.8–4.13:** Error handling (unmatched path with no fallback → 404, internal errors), response headers, status codes

All tests passing: `cabal test all` green (1723 examples, 0 failures, 6 pending; StaticAssets 43/43).

## Hlint

Hlint clean.

## Checklist

- [x] ADR approved (phase 3)
- [x] Security findings grounded (phase 12)
- [x] Performance findings grounded (phase 13)
- [x] Tests passing (phase 15)
- Hlint: clean ✓

## Known follow-up (non-blocking)

The per-request `listDirectory` case-sensitivity check (perf-11) is pure overhead on Linux (deploy target). A future PR can gate it behind a macOS-only flag (System.Info.os == darwin) to remove the overhead on the primary deployment platform. This does not block v1 — it is documented and tested but kept as-is to ship the feature now and optimize after profiling real-world usage.

Closes #707

## Summary by CodeRabbit

* **New Features**
  * Added optional static asset serving for web transport, including SPA deep-link fallback and cache-aware responses.
  * Added a new application-level configuration path to enable static assets alongside existing web routes.

* **Bug Fixes**
  * Improved path handling and security checks to prevent directory traversal and keep asset lookups safely within the configured root.
  * Static assets now return sensible 404s when files or fallback targets are unavailable.

* **Tests**
  * Added broad coverage for file serving, fallback routing, content types, cache behavior, and edge cases.

### DIFF
--- a/core/service/Service/Application/Transports.hs
+++ b/core/service/Service/Application/Transports.hs
@@ -16,7 +16,7 @@ import Maybe (Maybe (..))
 import Maybe qualified
 import Service.ServiceDefinition.Core (TransportValue (..))
 import Service.Transport (Transport (..), QueryEndpointHandler, EndpointHandler, Endpoints (..), EndpointSchema (..))
-import Service.Transport.Web (WebTransport (..), AuthEnabled, OAuth2Config, FileUploadEnabled (..), CorsConfig, HealthCheckConfig, IntegrationStatus)
+import Service.Transport.Web (WebTransport (..), AuthEnabled, OAuth2Config, FileUploadEnabled (..), CorsConfig, HealthCheckConfig, IntegrationStatus, StaticAssets)
 import Service.Transport.Web.Readiness (ReadinessConfig)
 import Service.Application.Types (ApiInfo)
 import Service.Query.Subscriber (QuerySubscriber)
@@ -58,10 +58,12 @@ runTransports ::
   -- ^ Optional integration selection status for /health reporting
   Maybe ReadinessConfig ->
   -- ^ Optional readiness endpoint configuration for WebTransport
+  Maybe StaticAssets ->
+  -- ^ Optional static SPA asset serving configuration for WebTransport (ADR-0066)
   QuerySubscriber ->
   -- ^ The live subscriber whose readiness state /ready will reflect
   Task Text Unit
-runTransports transportsMap endpointsByTransport schemasByTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig subscriber = do
+runTransports transportsMap endpointsByTransport schemasByTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig maybeStaticAssets subscriber = do
   transportsMap
     |> Map.entries
     |> Task.forEach \(transportName, transportVal) -> do
@@ -84,7 +86,7 @@ runTransports transportsMap endpointsByTransport schemasByTransport queryEndpoin
 
         -- Handle WebTransport specially to configure auth, OAuth2, and file uploads
         case transportName of
-          "WebTransport" -> runWebTransport transportVal commandEndpointsForTransport commandSchemasForTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig subscriber
+          "WebTransport" -> runWebTransport transportVal commandEndpointsForTransport commandSchemasForTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig maybeStaticAssets subscriber
           _ -> runGenericTransport transportVal commandEndpointsForTransport commandSchemasForTransport queryEndpoints querySchemas
 
 
@@ -110,9 +112,10 @@ runWebTransport ::
   Maybe HealthCheckConfig ->
   Maybe IntegrationStatus ->
   Maybe ReadinessConfig ->
+  Maybe StaticAssets ->
   QuerySubscriber ->
   Task Text Unit
-runWebTransport transportVal commandEndpoints commandSchemas queryEndpoints querySchemas maybeAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig subscriber = do
+runWebTransport transportVal commandEndpoints commandSchemas queryEndpoints querySchemas maybeAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig maybeStaticAssets subscriber = do
   case transportVal of
     TransportValue transport -> do
       -- Cast the existentially-typed transport to WebTransport
@@ -136,6 +139,7 @@ runWebTransport transportVal commandEndpoints commandSchemas queryEndpoints quer
             , integrationStatus = maybeIntegrationStatus
             , readinessConfig = maybeReadinessConfig
             , readinessProbe = Just (Subscriber.readinessOf subscriber)
+            , staticAssets = maybeStaticAssets
             }
       -- Build endpoints with the configured transport
       let endpoints :: Endpoints WebTransport =
--- a/core/service/Service/Transport/Web.hs
+++ b/core/service/Service/Transport/Web.hs
@@ -6,6 +6,9 @@ module Service.Transport.Web (
   CorsConfig (..),
   HealthCheckConfig (..),
   IntegrationStatus (..),
+  -- | Static asset serving (ADR-0066)
+  StaticAssets (..),
+  serveStaticAsset,
   server,
   isHealthCheckPath,
   buildHealthResponse,
@@ -34,9 +37,12 @@ import Log qualified
 import Data.Aeson qualified as Aeson
 import Data.ByteString qualified as GhcBS
 import Data.ByteString.Lazy qualified as GhcLBS
+import Data.Char (isDigit)
 import Data.IORef qualified as GhcIORef
 import Data.List qualified as GhcList
 import Data.Yaml qualified as Yaml
+import System.Directory qualified as GhcDir
+import System.FilePath qualified as GhcFilePath
 import GHC.TypeLits qualified as GHC
 import Json qualified
 import LinkedList qualified
@@ -145,6 +151,32 @@ data IntegrationStatus = IntegrationStatus
   }
 
 
+-- | Static asset serving configuration for WebTransport (ADR-0066).
+--
+-- When set, unmatched non-API GET requests are served from 'root' on disk,
+-- with correct Cache-Control and an optional SPA deep-link fallback.
+-- Set via 'Application.withStaticAssets'.
+--
+-- Example:
+--
+-- @
+-- StaticAssets
+--   { root = "static"                 -- the folder your build outputs to
+--   , spaFallback = Just "index.html" -- deep links fall back to the SPA shell
+--   }
+-- @
+--
+data StaticAssets = StaticAssets
+  { root :: Text
+  -- ^ Directory path (relative or absolute) whose contents are served.
+  -- Must be an existing, non-empty, readable directory at application startup.
+  , spaFallback :: Maybe Text
+  -- ^ Document served for unmatched paths (e.g., @Just "index.html"@) so
+  -- client-side routing can handle deep links. @Nothing@ → plain 404.
+  }
+  deriving (Show)
+
+
 -- | HTTP/JSON transport using WAI/Warp.
 data WebTransport = WebTransport
   { port :: Int,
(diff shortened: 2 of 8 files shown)


core/service/Service/Application/Transports.hs
```diff
@@ -58,10 +58,12 @@ runTransports ::
   -- ^ Optional integration selection status for /health reporting
   Maybe ReadinessConfig ->
   -- ^ Optional readiness endpoint configuration for WebTransport
+  Maybe StaticAssets ->
+  -- ^ Optional static SPA asset serving configuration for WebTransport (ADR-0066)
   QuerySubscriber ->
   -- ^ The live subscriber whose readiness state /ready will reflect
   Task Text Unit
-runTransports transportsMap endpointsByTransport schemasByTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig subscriber = do
+runTransports transportsMap endpointsByTransport schemasByTransport queryEndpoints querySchemas maybeWebAuth maybeOAuth2 maybeFileUpload maybeApiInfo maybeCors maybeHealthCheck maybeIntegrationStatus maybeReadinessConfig maybeStaticAssets subscriber = do
```
NickSeagull, line 66: Create a github issue to make these parameters to actually be a record like transport config or something, it is growing out of control

core/service/Service/Transport/Web.hs
```diff
@@ -235,9 +271,9 @@ readBodyWithLimit maxSize request = Task.fromIO do
 
       drainBody = do
         chunk <- Wai.getRequestBodyChunk request
-        if GhcBS.null chunk
-          then pure ()
-          else drainBody
+        case GhcBS.null chunk of
+          True -> pure ()
+          False -> drainBody
```
NickSeagull, line 276: roll this back, use if-then-else

core/service/Service/Transport/Web.hs
```diff
@@ -1150,9 +1188,8 @@ isOriginAllowed cors origin = do
   let origins = cors.allowedOrigins
   let lowerOrigin = origin |> Text.toLower
   -- Check for wildcard
-  if Array.contains "*" origins
-    then True
-    else origins |> Array.any (\o -> Text.toLower o == lowerOrigin)
+  Array.contains "*" origins
+    || (origins |> Array.any (\o -> Text.toLower o == lowerOrigin))
```
NickSeagull, line 1192: make this more readable

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+  StaticAssets ->
+  Wai.Request ->
+  (Wai.Response -> Task Text Wai.ResponseReceived) ->
+  Task Text Wai.ResponseReceived
+serveStaticAsset assets request respond = do
+  let respond404 = respond (Wai.responseLBS HTTP.status404 [] "")
+  -- Only serve GET; everything else → 404
+  case Wai.requestMethod request == "GET" of
+    False -> respond404
+    True -> do
```
NickSeagull, line 1392: I see you use a lot of case..of..True..False. This is an anti-pattern, use if-then-else. Modify your config/skills/knowledge so this doesn't happen

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+              |> Wai.rawPathInfo
+              |> Bytes.fromLegacy
+              |> Text.fromBytes
+              |> Text.toLinkedList
+              |> GhcFilePath.normalise
+      let pathSegments = GhcFilePath.splitDirectories rawPath
+      let hasDotDot = pathSegments |> GhcList.any (\s -> s == "..")
+      case hasDotDot of
+        True -> respond404
+        False -> do
```
NickSeagull, line 1405: another one

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+          let rootStr = assets.root |> Text.toLinkedList
+          let relPath = case rawPath of
+                '/' : rest -> rest
+                other -> other
+          -- Root path "/" has no relative component — serve SPA fallback directly.
+          case relPath == "" of
+            True -> serveStaticSpaFallback assets rootStr respond404 respond
+            False -> do
```
NickSeagull, line 1413: and another one

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+                          -- Case-sensitive guard: on case-insensitive file systems
+                          -- doesFileExist may return True for a differently-cased name.
+                          -- Verify the exact filename appears in the directory listing.
+                          let parentDir = GhcFilePath.takeDirectory fullPath
+                          let reqFileName = GhcFilePath.takeFileName fullPath
+                          dirContents <- GhcDir.listDirectory parentDir |> Task.fromIO
+                          case reqFileName `GhcList.elem` dirContents of
+                            False -> respond404
+                            True -> serveStaticFile fullPath respond
+
```
NickSeagull, line 1436: after reading all this function (serveStaticAsset) you are doing a pyramid-of-doom. Learn this new pattern: sentinels. Add it to your knowledge so when you face these kinds of validation, and especially in functions that return Task. You instead do the opposite check. E.g.

Instead of

```hs
if cond
  then do
    code1
    if cond2
      then do
        code2
        if cond3
          then XXX
          else respond404
      else
        respond404
    else
      respond404
   else
     respond404
```

you do this instead:

```hs
Task.unless (cond) respond404
code1
Task.unless (cond2) respond404
code2
Task.unless (cond3) respond404
XXX
```

This keeps the reading complexity down to a minimum by minimizing the indentation levels. Save this to memory. There is a lot of antipatterns like these everywhere, so we have to apply the boy-scout-principle too (when we find something like this in the file we are modifying we fix it (NOT ALL, only the ones we touch) )

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+        False -> do
+          let fallbackPath = GhcFilePath.normalise (rootStr GhcFilePath.</> fallbackDocStr)
+          let rootNorm = GhcFilePath.normalise rootStr
+          case GhcList.isPrefixOf rootNorm fallbackPath || fallbackPath == rootNorm of
+            False -> respond404
+            True -> do
+              fallbackExists <- GhcDir.doesFileExist fallbackPath |> Task.fromIO
+              case fallbackExists of
+                False -> respond404
+                True -> serveStaticFile fallbackPath respond
```
NickSeagull, line 1477: See? Another one

core/service/Service/Transport/Web.hs
```diff
@@ -1330,3 +1365,190 @@ unauthorizedResponse authError =
...
+      mimeBase = staticMimeBase ext
+      isOctetStream = mimeBase == "application/octet-stream"
+      isHashed = staticIsHashedAsset fileName
+      immutable = isHashed || isOctetStream
+      contentType = case immutable of
+        True  -> mimeBase
+        False -> staticMimeWithCharset ext
+      cacheControl = case immutable of
+        True  -> "public, max-age=31536000, immutable"
+        False -> "no-cache, must-revalidate"
```
NickSeagull, line 1496: It truly worries me that this implementation has been 100% vanilla haskell hallucination. Absolutely no use of nhcore, usage of let..in. Etc....

Issue #715, opened 2026-07-07
Continuous Generation pipeline — implementation tracker


NickSeagull:
Single tracking issue for implementing the Continuous Generation → Continuous Verification → Continuous Deployment pipeline.

**Goal:** PRs go from 2–3h-and-wrong to **≤30 minutes and correct**.

**Plan:** [docs/plans/2026-07-07-continuous-generation-pipeline-plan.md](https://github.com/neohaskell/NeoHaskell/blob/main/docs/plans/2026-07-07-continuous-generation-pipeline-plan.md) (lands via #714)

## Phases

- [x] **Phase 0 — Archive & truth reset**: full sweep of rotten AI artifacts (docs, all 14 skills, opencode configs) into `docs/archive/` with ARCHIVAL banners; verified CLAUDE.md stub + `codemap/` skeleton in the same atomic PR
- [x] **Phase 1 — Fast inner loop + telemetry**: profile, `-O0` dev flavor, ghcid `--outputfile` loop, link-free `--match` tests, worktree pool + last-good `dist-newstyle` bake/restore, frozen telemetry schema, golden-task accumulation (incl. transcripts)
- [x] **Phase 2 — Dialect enforcement**: dialect-first `.hlint.yaml` + missing CI step, edit-time deny hook, copy-adapt rule, vanilla-exception registry with rule-of-three promotion
- [x] **Phase 3 — Localization assets (`codemap/`)**: capabilities + extension-points YAML (with risk tags), committed signatures + sync check, module/function doc-ratchet, hiedb wrappers, routing smoke set
- [x] **Phase 4 — API knowledge delivery**: frequency-ranked hot card, doctest-mined phrasebook, local hoogle, plan-time symbol resolution, invented-API telemetry
- [x] **Phase 5 — Spec gate + verification**: contract-delta spec template (criteria with declared test levels, mechanically-triggered ADR), draft-PR gate + authorized continue signal, risk-tiered design-time security/perf review, test-first ordering, spec-drift check, failure policy, nightly non-blocking benchmarks
- [x] **Phase 6 — Release tail + learning loop**: criteria-based definition-of-done gate, revert drill, mechanical breaking-change policy, failure→asset-delta protocol, retrospective miner, weekly review, archive sunset

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #718, opened 2026-07-07
Phase 2: dialect enforcement — hlint inversion, edit hook, implementer skill

NickSeagull:
Part of #715 — **Phase 2 — Dialect enforcement** of the [pipeline plan](https://github.com/neohaskell/NeoHaskell/blob/main/docs/plans/2026-07-07-continuous-generation-pipeline-plan.md).

Goal: vanilla Haskell cannot survive to CI, with feedback in **milliseconds** (hook) → **seconds** (hlint) → build (GHC), instead of being discovered at review time (#708-class failures).

## Exit criteria — all verified live

- ✅ Probe edit with `foo $ bar` rejected by the hook locally (17/17 hook unit cases)
- ✅ Probe file with vanilla import + `$` fails hlint with teaching messages
- ✅ `./dev lint` green on the whole repo **with zero source changes** (current usage grandfathered)
- ✅ Hook wired in `settings.json`; CI lint job added

## Layer 1 — `.hlint.yaml` inverted to dialect-first

- **`modules:` restrictions**: Data.Text/Char/List/Map/Maybe/Either/Set/Vector, System.IO/Environment/Directory, Data.Aeson, GHC.Base → allowed only in their Core wrappers + **129 grandfathered `within:` entries generated from actual imports** (2026-07-07). Current usage stays legal; new vanilla imports fail with "Use Core `X`" messages.
- **`functions:` restrictions**: `fromJust` banned outright; `undefined` grandfathered to 5 modules. `error` deliberately **not** restricted — triage showed hlint matches bare names, colliding 101× with `Log.error`/record fields; `NoImplicitPrelude` + the hook cover it.
- **Teaching hint**: `a $ b` → `b |> a`.
- **21 default hints ignored, each with a written triage verdict** — the headline finding: hlint's defaults (`Use lambda-case`, `Replace case with fromMaybe`, `Use isDigit`, Bifunctor/TupleSections hints…) actively coach code *toward* vanilla idioms. The old config was anti-dialect guidance in a repair loop.
- **Escape hatch in the config header**: `within:` + justification + `belongs-in:` note; rule of three promotes a Core primitive.

## Layer 2 — edit-time hook (`dialect-guard.py`, ~50ms)

Rejects `$`, `where` clauses, `Either`, `pure`/`return` usage, vanilla + unqualified open imports, `case`-of-Bool. Precision engineering so it doesn't cry wolf: **only added lines** (old/new diff) are checked, comment lines skipped, `module/class/instance/data … where` allowed, TH splices allowed, `pure = …` instance definitions allowed, `docs/archive/` skipped, `-- HOOK-ALLOW: <reason>` escape for false positives. Every rejection quotes the rule and the escape hatch.

## Layer 3 — CI gate

New `lint` job in `checks.yml`: pinned hlint 3.10 binary (sha256-verified, matches the dev-shell version), same config as `./dev lint` (new verb, doctor-registered) — local pass ⇒ CI pass.

## Also

- **`neohaskell-implementer` skill rebuilt** from Phase 0 salvage: copy-adapt discipline (no `.hs` from a blank buffer), the `./dev` repair protocol, escape-hatch teaching, 2-round circuit breaker.
- **Doctor extended**: skills are agent-visible docs, so per the governing rule they get a gate — every `./dev <verb>` referenced by a SKILL.md must be registered (adversarially verified).

## Summary by CodeRabbit

* **New Features**
  * Added a `lint` command and CI “dialect gate,” including a dedicated CI lint job.
  * Introduced a pre-edit dialect validation hook that blocks specific vanilla Haskell constructs on newly added lines.
* **Documentation**
  * Expanded contributor workflow and dialect rule guidance; updated README and agent policy to reflect the new enforcement phases and escape-hatch process.
* **Bug Fixes**
  * Strengthened governance via `doctor` with verification of contributor workflow references and the hook’s self-test.
* **Tests**
  * Added automated coverage scenarios for the dialect validation hook, including allow-list behavior.
* **Chores**
  * Replaced `.hlint.yaml` with a comprehensive dialect-first configuration and updated CI to use a pinned `hlint` binary.


NickSeagull:
Bot review addressed (commit above): **all 5 findings valid, all fixed.**

The Critical was a genuine catch: the hook now reads the **official stdin JSON contract** (`{tool_name, tool_input}`) with env-var fallback — and it implies the repo's pre-existing fourmolu PostToolUse hook (env-var based) may be silently dead; flagged for separate verification rather than churned here. Also fixed: `pure x = …` instance definitions no longer flagged (lookahead exempts arg-ful definitions, still catches `==` comparisons), lint path list deduplicated (CI reuses `./dev lint` via `HLINT=` override), README example aligned, skill enforcement claim made honest.

Re the design question raised alongside this review — *why a Python hook instead of hlint?* — panel-reviewed answer now embedded in the hook's header: **hlint is the law, the hook is the teacher.** The hook runs before the file exists (PreToolUse sees a fragment; hlint needs a parseable module), tolerates mid-edit broken states, covers rules hlint can't express (`where` syntax, `Either` as a type, `pure` usage-vs-definition), and needs no toolchain (~50ms on cold cloud sessions). Every hook rule now names its canonical gate; on disagreement the law wins. Verified 20/20 unit cases in both payload modes.

NickSeagull:
Bot round 3: **2 new findings valid, both fixed** — (1) pattern-matching instance methods (`pure Nothing = …`, `pure (Just x) = …`, `pure [] = …`) no longer flagged; the definition exemption moved out of regex-lookahead into `exempt()` where it's testable per rule id; `pure x == y` still blocked. (2) `HOOK-ALLOW` for the case-of-Bool rule is now scoped to the matched construct's own lines — a stray allow elsewhere in the edit can't silence it (regression case added for exactly that). 33 cases, 7 rules fully covered, self-test + doctor green. The remaining 5 comments were re-posts of round-1 findings, verified already fixed in current code.

PR #719, opened 2026-07-07
Phase 3: codemap — localization becomes lookup, not search

NickSeagull:
Part of #715 — **Phase 3 — Localization assets** of the [pipeline plan](https://github.com/neohaskell/NeoHaskell/blob/main/docs/plans/2026-07-07-continuous-generation-pipeline-plan.md).

Goal: plan-time localization in ≤3 minutes with zero exploration — exact paths and symbols resolved while planning, written into the plan; execution never searches.

## Exit criteria — all met

| Criterion | Result |
|---|---|
| `check.py` green in CI | ✅ 321 exposed modules **owned exactly once**, 30 capabilities, 183 unique aliases |
| Signatures committed + sync-checked | ✅ 297 modules / 3327 exports in 4 grep-sized files; `codemap-sync` job in test.yml regenerates + diffs |
| who-calls returns capability-tagged output | ✅ `./dev who-calls yield Task` → *621 refs — auth: 115, testlib: 85, event-store: 83, …* |
| Routing smoke ≥8/10 | ✅ **10/10 by a fresh-context agent** (read only the two YAML files; requests grep'd without expected answers; all 3 vocabulary-gap probes correct, incl. "serve a SPA" → `http-transport`) |
| MANIFEST back-filled | ✅ archive destinations point at real files now |

## The assets (all generated-or-gated)

- **`capabilities.yaml`** — 30-capability ontology: owns-globs, responsibilities seeded from the Phase 0 salvage's verified claims, aliases as the intent-vocabulary bridge, `security-sensitive`/`perf-sensitive` tags that drive the Phase 5 review gates.
- **`extension-points.yaml`** — where things that don't exist yet go: 6 kinds (core primitive, command, query, transport feature, outbound integration, pipeline asset), each with create/register/tests/skill.
- **`check.py`** (new `codemap` job in checks.yml) — ownership exactly-once (adding a module without a map entry **fails the PR**), no ghost globs, alias uniqueness, extension paths + skills exist, **doc-ratchet may only decrease** (baseline recorded honestly: 194 undocumented modules, 1727 undocumented exports). The gate caught 12 defects in its own seed data on first run.
- **`signatures/*.txt`** — generated hoogle API surface; agents grep these instead of opening source files.
- **hiedb layer** — `-fwrite-ide-info` in cabal.project, hiedb in the dev shell, `./dev hiedb` / `who-calls` / `where-defined` with capability-joined output. Known caveat (documented in the localizer skill): record-field accessors may not resolve via name-refs.
- **`neohaskell-localizer` skill** — the plan-time routing contract (closed-list selection, signature grep, edit-set expansion, confidence rule: escalate, never guess).

## Also

`with-toolchain`'s stale-shell fast path is **removed** after its third failure (`bash`/`python3` exist on every PATH, so "command exists" proves nothing) — always `nix develop` now, ~0.4s warm. Correct beats fast.

New verbs: `codemap`, `codemap-check`, `hiedb`, `who-calls`, `where-defined` (13 total, doctor-gated).

## Summary by CodeRabbit

* **New Features**
  * Added a codemap-based localization workflow for finding code, related symbols, and extension points more quickly.
  * Added new developer commands and scripts for refreshing codemap data, rebuilding the symbol index, and tracing references or definitions.

* **Bug Fixes**
  * Added CI checks to keep codemap data and generated artifacts in sync.
  * Improved validation to catch routing ambiguities and out-of-date documentation counters earlier.

* **Documentation**
  * Expanded guidance for the new localization workflow and updated supporting references.


NickSeagull:
Review round addressed: **all 10 CodeRabbit findings valid and fixed**, plus Nick's testbed doctrine encoded (extension points renamed to `new-{command,query}-machinery` — user-app commands/queries live in user apps; testbed rows are demo + integration-test surface only). Standout fixes: the codemap CI job was bypassing the very toolchain doctrine it gates (bot quoted our own `with-toolchain` rationale back at us — now `PYTHON= ./dev codemap-check`, the sanctioned override pattern); `check.py` gained an all-files exactly-once check that immediately caught two real ownership overlaps beyond the one reported; `who-calls`/`where-defined` injection surface removed (argv end-to-end). All gates re-verified green.

NickSeagull:
codemap-sync root cause found and fixed: **haddock's hoogle output is not byte-stable across platforms** (Linux emits fewer doc comments than macOS — identical CI hashes across both failed runs proved Linux is internally deterministic, just different). Fix: signatures are now canonicalized to **pure API surface** (doc/blank lines stripped at generation — platform-stable by construction), and the doc-ratchet is **source-derived** (same exposed-modules resolution as check.py; new baseline 212 modules / 562 exports lands atomically with the method change). The sync check stays byte-exact — now the artifacts deserve it. Phase 4 will mine docs/doctests from source, not from these files.

PR #740, opened 2026-07-22
chore(dialect): enforce Task control flow — |> discard, Task.when, Task.unless

NickSeagull:
## Summary

Enforces three NeoHaskell Task-context control-flow idioms, each at the layer
that can express it precisely, and migrates the genuine violations already in
the tree. Spec: `docs/changes/002-task-control-flow-dialect-rules.md`.

| Vanilla shape | Dialect form | Enforcement layer |
|---|---|---|
| `_ <- action` | `action` piped into `discard` | edit hook (`no-discard-bind`) |
| `if c then pass else a` | `Task.unless c do a` | hlint rewrite |
| `if c then a else pass` | `Task.when c do a` | hlint rewrite |

## Why these layers

- **`_ <- action` → edit hook.** A discarded bind is a do-*statement*, not an
  expression, so hlint's `lhs/rhs` rewrites can't match it (verified: hlint
  emits no hint for `_ <- x`). The hook's added-lines-only scope grandfathers
  the **142** existing sites (~110 idiomatic parser combinators the request
  says not to churn) while teaching new code; `Syntax/**` / `NeoQL/**` are
  path-exempt. `discard` is generic (`Mappable`), so the rewrite is always valid.
- **`if … pass` → hlint.** hlint parses real Haskell, so it matches the full
  `if`-expression precisely — multi-line `do`-block branches, and it correctly
  excludes `case`-arm `-> pass`. The rules key on the **literal** `pass`
  (multi-char = literal in hlint hints; single-char `c`/`a` = wildcards), so
  they fire on dialect Task code and not on `pure ()`/`return ()` legacy or
  non-Task Applicatives. GHC (`Task.when`/`unless` are Task-typed) + the hlint
  escape hatch backstop the one residual non-Task `pass` case.

## Ignore replacement

The blanket `Use when` / `Use unless` ignores only muted the *vanilla*
`Control.Monad` suggestion; they're removed and replaced by the exact Task
rules. The three non-Task boundary sites the un-muted builtin re-exposes
(`Console` IO, `Service.Transport.Web` WAI, `Service.OutboundIntegration.TH`
`Q`-monad) get a narrow, justified module-scoped ignore.

## Migration

All **29** genuine Task sites the new rules flag are migrated — **21**
`if c then a else pass` → `Task.when`, **8** `if c then pass else a` →
`Task.unless` (including multi-line branches hlint catches structurally). Each
is behaviour-identical. Non-Task sites left untouched; existing `_ <-` discards
grandfathered.

## Also in this PR — pipeline approval-flow doc (no `.github/` changes)

`.claude/skills/neohaskell-pipeline/SKILL.md` is clarified so the **server-local
session is the canonical Gate approval flow**: the orchestrator runs in a
persistent local tmux session, approval arrives out-of-band (RAMSYS / Discord /
direct local interaction) and is recorded with `./dev pipeline approve <gate>
--by <who> --via <channel>`, which resumes the same session in place;
`.pipeline/state.json` is the authorization record. A GitHub comment is optional
communication only — never the resume mechanism. **This PR makes no changes
under `.github/`.**

## Criteria → proof

| ID | Proof |
|---|---|
| C1 / C2 | `dialect-guard.py --self-test` — blocking + passing cases for `no-discard-bind` |
| C3 / C4 | `./dev lint` fires `NeoHaskell: use Task.when` / `use Task.unless` on the canonical forms |
| C5 | 29 sites migrated; `./dev lint` clean; `nhcore-test-service` green |
| C6 | `./dev doctor` OK |

## Verification (all green locally)

- `./dev lint` (core integrations testbed) — **No hints**
- `dialect-guard.py --self-test` — OK (37 cases, 8 rules); `./dev doctor` — OK
- `lib:nhcore` typechecks; `nhcore-test-service` — **1062 examples, 0 failures**
- spec-check / spec-drift / changelog — all pass

## Gate

Gate 1 (spec approval) was granted by the maintainer out-of-band and recorded
locally in `.pipeline/state.json` (the authorization record). This PR is
**ready for review**, awaiting **Gate 2** — the maintainer's final review +
merge. Not merged.

## Summary by CodeRabbit

* **Enhancements**
  * Standardized Task control-flow patterns across validation, storage, uploads, event processing, and integration workflows.
  * Added guidance to express discarded actions with the recommended pipeline syntax.
  * Added automated checks and lint suggestions to encourage consistent `Task.when` and `Task.unless` usage.

* **Documentation**
  * Added release documentation describing the control-flow conventions, validation criteria, and expected source-level impact.

* **Bug Fixes**
  * Improved detection of unsupported discarded-bind patterns while preserving valid parser and comprehension syntax.


NickSeagull:
Thanks — triaged all three:

**1. `no-discard-bind` misses the multiline form** — ✅ fixed in `6814385`. Broadened the regex to `^\s*_\s*<-\s*(\S|$)` so `_ <-` with the action on an indented continuation line is caught (the `_ <-` line is added on its own). Added a blocking regression case (`multiline discarded bind blocked`).

**2. Parser exemption too broad** — ✅ fixed in `6814385`. Restricted the exemption to `core/syntax/` / `core/neoql/` (was any `/syntax/` or `/neoql/` segment). Added a blocking regression case for a near-miss dir (`core/syntaxtree/`), which now stays blocked; the real parser packages remain exempt. self-test: 39 cases, 8 rules.

**3. `/revert` kill switch on every trigger path (`claude.yml` + `check-claude-trigger`)** — out of scope for this PR: those files are no longer part of it. The GitHub-Actions approval-trigger work was reverted by design — the orchestrator resumes via its persistent local session, not GitHub Actions, so this PR makes **no `.github/` changes** and adds no trigger checker. Your observation is valid as a *pre-existing* asymmetry in `main`'s `claude.yml` (the `/revert` guard is only on the `issue_comment` branch), but it should be handled in a separate change rather than reintroduced here.

NickSeagull:
Re the `runResult` hot-card finding — this is a false positive, no change made:

- The entry is in **`codemap/api-hot.md`**, not `CHANGELOG.md` (the comment is anchored to the wrong file).
- `codemap/api-hot.md` is **deterministically generated** by `./dev codemap` — a frequency-ranked *top-N* card. This PR's **29** new `Task.when`/`Task.unless` call sites push `Task.when` (29 sites) into the card, displacing `Task.runResult` (13 sites) below the cutoff. That is the correct, expected output.
- Restoring `runResult` by hand would make the committed card diverge from `./dev codemap` output and **fail the required `codemap-sync` CI gate** — which currently passes precisely because the card matches the regenerated one.

Merging per maintainer Gate-2 approval.

PR #748, opened 2026-08-01
feat(crypto): HMAC-SHA256 signWith/verifyWith with opaque HmacKey

stepandra:
## What

New public `Crypto` module in nhcore: general-purpose HMAC-SHA256 signing with constant-time verification — the primitive for webhook body authentication (sender signs the raw request body, receiver recomputes and compares).

Data-last, pipe-friendly, per the dialect rules:

```haskell
someBytes
  |> Crypto.signWith myHmacKey

someBytes
  |> Crypto.verifyWith myHmacKey incomingSignature
```

## Why

Discussed with @NickSeagull (chat, 2026-08-01): applications should not pull crypto packages from Hackage directly; `Crypto.sign` belongs in nhcore. The only HMAC-SHA256 in nhcore today is internal to `Auth.OAuth2.StateToken`, shaped around a fixed `StatePayload` + TTL — no way to sign an arbitrary body. This also provides the primitive for the inbound half anticipated by ADR-0058 ("public webhook receiver… signed with HMAC at the body level rather than JWT") without deciding transport policy.

## API

```haskell
data HmacKey                                      -- opaque, redacted Show
hmacKeyFromText  :: Text -> Result Text HmacKey    -- >= 32 bytes enforced
hmacKeyFromBytes :: Bytes -> Result Text HmacKey
generateHmacKey  :: Task err HmacKey               -- 256-bit random key
signWith         :: HmacKey -> Bytes -> Text       -- 64-char lowercase hex
verifyWith       :: HmacKey -> Text -> Bytes -> Bool  -- constant-time
```

Design notes:

- **Pure `signWith`** (not `Task`): HMAC is deterministic; purity keeps the pipeline ergonomic. Effects stay in `generateHmacKey` only.
- **Wire format**: lowercase hex (GitHub/Stripe-style header convention); `verifyWith` accepts uppercase input, compares in constant time via `BA.constEq`.
- **`Auth.OAuth2.StateToken` untouched**: migrating its private `HmacKey` onto `Crypto.HmacKey` is a possible follow-up refactor, deliberately out of scope to avoid touching auth behavior here (noted in the change spec).

## Validation

- `docs/changes/003-crypto-hmac-sign-verify.md` (spec-first commit); `./dev spec-check` OK, `./dev spec-drift` OK — generated surface honors the contract delta
- 17 new `CryptoSpec` examples incl. independently computed known-answer vectors, tamper/wrong-key/truncation/non-hex rejection, redaction, key-length boundaries; full `nhcore-test-core`: **1078 examples, 0 failures** (3 pre-existing pending)
- hlint 3.10 (dialect config): no hints; `./dev codemap-check` OK (signatures regenerated); CHANGELOG entry appended via `./dev changelog --append`
- Validated on host GHC 9.8.4 without Nix (`--index-state=2025-10-31T00:25:59Z`) — flagging in case the Nix toolchain surfaces drift

## Summary by CodeRabbit

* **New Features**
  * Added HMAC-SHA256 key creation, parsing, and secure random key generation.
  * Added signing and constant-time signature verification.
  * Signatures use lowercase hexadecimal; verification accepts mixed-case input.
  * Keys require at least 32 bytes and are redacted when displayed.
  * Added secure random byte generation, including support for empty results.

* **Documentation**
  * Added documentation describing the cryptographic APIs, security requirements, and behavior.


core/core/Crypto.hs
```diff
@@ -0,0 +1,157 @@
...
+  hmacKeyFromBytes (Text.toBytes secret)
+
+
+-- | Create an HMAC key from raw secret bytes.
+--
+-- The secret must be at least 32 bytes (256 bits) for security.
+hmacKeyFromBytes :: Bytes -> Result Text HmacKey
+hmacKeyFromBytes secret = do
+  let secretBytes = Bytes.unwrap secret
+  let len = BS.length secretBytes
```
NickSeagull: Can you add length to the Bytes module if it's not there and use that instead of unwrapping?

core/core/Crypto.hs
```diff
@@ -0,0 +1,157 @@
...
+--
+-- Creates a 32-byte (256-bit) key suitable for HMAC-SHA256.
+--
+-- WARNING: In production, load a persistent key from environment/config
+-- using 'hmacKeyFromText' instead. A key generated at runtime cannot
+-- verify signatures issued before a restart, and no other party can
+-- verify signatures made with it.
+generateHmacKey :: Task err HmacKey
+generateHmacKey = Task.fromIO do
+  randomBytes <- Random.getRandomBytes 32
```
NickSeagull: Same with this, maybe Bytes.getRandom? IIRC, there is something already for getting random Ints in core, we should do a similar API

NickSeagull (commented):
Thanks a lot for the PR! 

Everything looks good, just left two small requests to improve the ergonomics of future PRs that make heavy usage of Bytes like this one.

Looking forward to merging this!

core/core/Crypto.hs
stepandra: Done in 696dbb6 — `Bytes.length` already existed in the module, so `hmacKeyFromBytes` now validates through it instead of unwrapping.

core/core/Crypto.hs
stepandra: Added `Bytes.getRandom` in 696dbb6, mirroring the `Int.getRandom` API: `getRandom :: Int -> Task _ Bytes` (cryptographically secure — it wraps the OS random source, not the `System.Random` PRNG, since Bytes randomness is typically for secrets). `Crypto.generateHmacKey` now just does `Bytes.getRandom 32`.

One structural note: `Bytes` could not import `Task` directly (`Task` → `Text` → `Bytes` cycle), so the newtype moved to an internal `Bytes.Internal` module that `Text` and `Bytes` both import. Public API is unchanged — `Bytes (..)` is still re-exported from `Bytes`. Happy to reshape if you prefer a different cycle-break.

stepandra: Follow-up in c80e0d7: `Bytes.Internal` is now hidden (`other-modules`, matching the existing `Parser.Internal` precedent) — it exists purely as the cycle-break and is not importable by applications. Public API delta is just `Bytes.getRandom`.

NickSeagull (approved):

NickSeagull:
@stepandra can you run ./dev code map once again and commit the result pls?

stepandra:
Done — regenerated the codemap and committed (8e90da4). The drift was the five Bytes instance signature lines that disappeared from the public surface after hiding Bytes.Internal.

CHANGELOG.md
```diff
@@ -13,6 +13,29 @@ has been cut yet — everything accrues under Unreleased until the first tag.)
 
 ## [Unreleased]
 
+### 004-crypto-hmac-sign-verify — Change 003: Add Crypto module with HMAC-SHA256 signWith/verifyWith
```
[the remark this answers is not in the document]
NickSeagull: Fixed in 6837798: the spec is now `004-crypto-hmac-sign-verify.md`, both generated headings say `Change 004`, and `./dev changelog --check origin/main` passes. I also compared the committed CHANGELOG block byte-for-byte with `./dev changelog docs/changes/004-crypto-hmac-sign-verify.md`; they match.

Issue #798, opened 2026-08-16
Rename one-shot Agent integration to CommandRouter and make command descriptors provider-neutral


NickSeagull:
## Summary

Rename the existing one-shot `Integration.Agent` abstraction to `Integration.CommandRouter` and extract a provider-neutral descriptor for NeoHaskell commands exposed to language models.

The current implementation performs one bounded operation:

```text
prompt + allowed commands
  -> one OpenRouter request
  -> one selected CommandPayload
  -> stop
```

It has no logical session, conversation state, owned process, lifecycle, or recovery behavior. `Agent` therefore overstates its semantics and prevents the name from being used precisely for a future persistent-agent abstraction.

This issue follows the implemented direction originally proposed in #367, but corrects the public vocabulary and provider boundary.

### Rename the one-shot surface

```haskell
CommandRouter.route
  prompt
  [ Ai.command @CreateBugReport
  , Ai.command @CreateFeatureRequest
  ]
  model
  onError
```

Choose the final shared descriptor module name during implementation, but preserve this provider-neutral data:

```text
command name        <- NameOf
command description <- Documented
input JSON Schema   <- ToSchema
```

### Move provider rendering into adapters

The current `CommandTool.toolDefinition` contains a pre-rendered OpenRouter function wrapper. Replace it with the raw provider-neutral input schema.

```text
provider-neutral command descriptor
  -> OpenRouter adapter renders a function tool
  -> future backends render their own protocol shape
```

## Requirements

- Preserve the current one-shot command-routing behavior.
- Preserve validation that the selected command belongs to the explicitly supplied catalogue.
- Preserve typed command decoding and normal `CommandPayload` dispatch.
- Keep prompts, schemas, and provider responses out of logs.
- Update module names, tests, examples, ADR-0045, package exports, and migration notes coherently.
- Decide and document whether a temporary compatibility re-export is warranted; do not maintain two competing canonical names indefinitely.

## Non-goals

- Persistent sessions or multi-turn conversation state.
- Child-process lifecycle.
- Claude Code support.
- Durable integration execution streams.
- Synchronous model tools beyond the current one-shot routing contract.

## Acceptance criteria

- [ ] The canonical public name describes one-shot command routing rather than a persistent agent.
- [ ] The shared command descriptor contains no OpenRouter-specific wrapper fields.
- [ ] OpenRouter rendering occurs only inside the OpenRouter/CommandRouter adapter boundary.
[...]

PR #802, opened 2026-08-16
spec: make NameOf the single type identity and add the Descriptor primitive (#801)

### LINKED ISSUE #801: Make NameOf the single type identity: drop Documented.name and add a name+description+schema descriptor primitive
## Summary

Make `NameOf` the single identity of a type, remove `Documented.name`, and introduce the reusable descriptor primitive that ADR-0013 specified but never implemented.

Today "a thing that carries a name, a description and a schema" exists four times in partial, incompatible shapes:

| Shape | Where | Carries |
|---|---|---|
| `Schema.FieldSchema` | `core/schema/` | field name, schema, required, description |
| `Service.Transport.EndpointSchema` | `core/service/` | request/response schema, description, deprecated |
| `QueryDefinition` | `core/service/` | name (`Text`), schema |
| `Integration.Agent.CommandTool` | `integrations/` | tool name, description, pre-rendered OpenRouter JSON |

ADR-0013 section 6 already planned the shared version (`commandSchema :: Schema` plus `commandDocumentation :: Maybe DocumentedInfo { docName, docDescription, docDeprecated }`), captured at `Service.command` time. `DocumentedInfo` was never implemented — zero occurrences in the codebase. `Documented` shipped as the leftover shell of that plan.

## The identity problem

Two names exist for the same type:

- `NameOf` — `type family NameOf (t :: Type) :: Symbol` (`core/service/Service/Command/Core.hs:63`). Type-level, so GHC can match on it; it drives `emitCommand`, `makeCommandPayload`, `buildHandler`, `withTransport` and the `KnownHash` command registry. Generated by the TH markers as the literal type name (`core/service/Service/CommandExecutor/TH.hs:490`), overridable by hand.
- `Documented.name` — class method defaulting to `TypeName.reflect @value` (`core/traits/Documented.hs`), user-overridable, value-level `Text`.

They already diverge. In `integrations/test/Integration/Agent/TestFixtures.hs` the type is `AddItemCommand` but `type instance NameOf AddItemCommand = "AddItem"`, so `Documented.name` yields "AddItemCommand" while dispatch expects "AddItem".

`Integration.Agent.commandTool` (`integrations/Integration/Agent.hs:91-93`) sidesteps this by taking the name from `NameOf` and only the description from `Documented` — it never calls `Documented.name`. Any future descriptor that naively reads `Documented.name` would emit tools whose names fail `validateToolName` and never dispatch: a runtime failure, invisible at compile time, limited to types whose `NameOf` differs from their type name.

`Documented.name` has **zero call sites** across `core/`, `testbed/` and `integrations/`. The only consumer of `Documented` at all is `Documented.description`, on that single line.

## Proposed direction

1. Remove `name` from the `Documented` class. `Documented` keeps `description`, `examples`, `deprecated` — human-facing text only, no identity.
2. `NameOf` becomes the single identity of a type. Where a non-concept type needs a display string, call `TypeName.reflect @value` directly, which is exactly what the deleted default did.
3. Add the provider-neutral descriptor primitive that bundles the triple:

```haskell
data Descriptor = Descriptor
  { name :: Text          -- from NameOf
  , description :: Text   -- from Documented
  , schema :: Schema      -- from ToSchema
  , examples :: Array Json.Value
  , deprecated :: Bool
  }

Descriptor.of ::
  forall value name.
  (Documented value, ToSchema value, NameOf value ~ name, KnownSymbol name) =>
  Descriptor
```

Requiring `KnownSymbol (NameOf value)` scopes the primitive to concept types (commands, queries, entities, transports): a type without a `NameOf` instance fails to compile, rather than silently getting a reflected name that cannot route.

`examples` is flattened to `Array Json.Value` via `ToJSON`, because `Documented.examples :: Array value` is type-indexed and cannot live in a monomorphic record. The flattened form is what both OpenApi example fields and AI few-shot prompts need.

4. The descriptor stays provider-neutral. Rendering lives outside it and downstream of it: `Schema.JsonSchema.toJsonSchema`, `Schema.OpenApi.toOpenApiSchema`, and the OpenRouter adapter in #798.

### Placement

`nhcore` is a single `library` with multiple `hs-source-dirs`, so a module under `core/schema/**` may depend on `Service.Command.Core` without a cabal change, and `Service/Command/Core.hs` imports neither `Core`, `Schema` nor `Documented`, so no import cycle forms. Caveat: re-exporting `Descriptor` from `Core.hs` would create `Core -> Descriptor -> Service.*`; either keep it out of the `Core` re-export, or place the module in the service layer.

## Requirements

- `Documented` no longer exposes `name`; the behavior of `Documented.description` is unchanged.
- A migration note documents the replacement: `type instance NameOf T = "..."` for identity, `TypeName.reflect @T` for a display string.
- `Descriptor.of` derives its name from `NameOf`, never from a value-level fallback.
- The descriptor contains no provider-specific or pre-rendered fields.
- Document the relationship to ADR-0013 section 6 (`DocumentedInfo` superseded by `Descriptor`).

## Non-goals

- Migrating `FieldSchema`, `EndpointSchema` or `QueryDefinition` onto `Descriptor`. That is a follow-up once the primitive exists.
- The `Integration.Agent` to `CommandRouter` rename and the OpenRouter adapter boundary (#798).
- Any change to `ToSchema` derivation or to the `Schema` ADT.

## Acceptance criteria

- [ ] `Documented` has no `name` member and the package builds with no call sites left behind.
- [ ] `Descriptor` and `Descriptor.of` exist, with the name sourced from `NameOf`.
- [ ] A type lacking a `NameOf` instance produces a compile error at `Descriptor.of`, covered by a test.
- [ ] `Descriptor.of` is covered for: description present, description empty, `NameOf` differing from the type name, and examples rendered to JSON.
- [ ] An ADR records the identity decision and supersedes the `DocumentedInfo` sketch in ADR-0013.
- [ ] Migration notes and package documentation updated.
- [ ] `./dev lint` and the full test suite pass.

## Breaking change

`Documented(..)` is re-exported from `Core.hs`, so any downstream app overriding `name` stops compiling. Migration: move the identity to `type instance NameOf`.

## Related

- #798 — CommandRouter rename; consumes this descriptor instead of the pre-rendered `CommandTool.toolDefinition`.
- ADR-0013 — introduced `Documented` (commit 42c57d5a, PR #312, issue #305) and specified the unimplemented `DocumentedInfo`.
- ADR-0045 — the Agent integration, the only current consumer of `Documented`.

NickSeagull:
**Gate 1 — spec only.** This PR's diff is the contract, not the implementation. Implementation starts once a maintainer approves.

Closes #801 (on merge of the implemented change).

## What is being proposed

- Remove `name` from the `Documented` class. `Documented` keeps `description`, `examples`, `deprecated` — human-facing text, no identity.
- `NameOf` becomes the single identity of a type. Display strings for non-concept types come from `TypeName.reflect @value`, which is exactly what the deleted default ran.
- Add `Descriptor` (`core/schema/Descriptor.hs`): the provider-neutral name + description + schema + examples + deprecated record, built by `Descriptor.describe @value`.

## Two deviations from the issue, both deliberate

1. **`Descriptor.describe`, not `Descriptor.of`** — `of` is a Haskell reserved word (`case … of`) and cannot be a function name. `Descriptor.reflect` / `Descriptor.forType` are the alternatives; renaming is a one-line change at this gate.
2. **`ToJSON value` is in the constraint set**, which the issue's sketch omits. `Documented.examples :: Array value` is type-indexed; flattening it to the record's `Array Json.Value` *is* a `ToJSON` call. Every concept type already has the instance.

## Why the compile-time guard matters

`KnownSymbol (NameOf value)` is required, so a type with no `NameOf` instance fails at the call site instead of silently getting a reflected name that cannot route. Without it, a descriptor that read the old `Documented.name` would emit tool names that fail `Integration.Agent.Internal.validateToolName` and never dispatch — a runtime failure, invisible at compile time, only for types whose `NameOf` differs from their type name (`AddItemCommand` / `"AddItem"` in the existing fixtures).

## Breaking

`Documented (..)` is re-exported from `Core`, so a downstream app overriding `name` stops compiling. Migration: `type instance NameOf T = "…"`. Inside this monorepo the blast radius is zero — `Documented.name` has no call sites and no overriding instances.

## Contents of this PR

| File | |
|---|---|
| `docs/changes/007-nameof-single-identity-descriptor.md` | the spec: contract delta + 6 criteria (all `unit`), each naming its proving test |
| `docs/decisions/0075-nameof-single-type-identity.md` | ADR (breaking-change trigger); supersedes the `DocumentedInfo` sketch in ADR-0013 §6 |

`./dev spec-check` and `./dev adr-check` are green. `./dev spec-check --plan` routes to **no** design reviews (`traits` and `schema` carry no risk tags); test impact is `core/test-core/**`.

## Not in scope

Migrating `FieldSchema` / `EndpointSchema` / `QueryDefinition` / `CommandTool` onto `Descriptor` (follow-up), the `Integration.Agent` → `CommandRouter` rename and OpenRouter adapter boundary (#798), and any change to `ToSchema` derivation or the `Schema` ADT.

### DIFF
--- a/docs/changes/007-nameof-single-identity-descriptor.md
+++ b/docs/changes/007-nameof-single-identity-descriptor.md
@@ -1,0 +1,157 @@
+# Change 007: Make `NameOf` the single type identity and add the `Descriptor` primitive
+
+Issue #801 asks for two things that are really one thing. First, a type must
+have **one** identity: today `NameOf` (type-level, drives `emitCommand`,
+`makeCommandPayload`, `buildHandler`, `withTransport` and the `KnownHash`
+command registry) and `Documented.name` (value-level `Text`, defaulting to
+`TypeName.reflect`) both claim to name a type, and they already disagree —
+`Integration.Agent.TestFixtures.AddItemCommand` has
+`type instance NameOf AddItemCommand = "AddItem"`, so `Documented.name` would
+answer `"AddItemCommand"` while dispatch expects `"AddItem"`. `Documented.name`
+has zero call sites in `core/`, `testbed/` and `integrations/`, and no instance
+anywhere overrides it, so it is a trap with no users: it is removed, and
+`NameOf` becomes the single identity. Second, the "name + description + schema"
+triple that ADR-0013 section 6 sketched as `DocumentedInfo` and never
+implemented is shipped for real as a provider-neutral `Descriptor`, whose name
+comes from `NameOf` — so a type that cannot route cannot get a descriptor
+either, at compile time rather than at runtime.
+
+```yaml spec
+issue: issue#801                # issue#NNN or adhoc:<slug>
+kind: refactor                  # feature | bug | refactor
+touches: [traits, schema]       # capability IDs from codemap/capabilities.yaml (closed list)
+breaking: true                  # MUST be true if the contract delta has any `-` line
+new-dependency: false           # any new build-depends / flake input
+new-capability: false           # this change adds a row to codemap/capabilities.yaml
+new-extension-point: false      # this change adds a row to codemap/extension-points.yaml
+```
+
+## Contract delta
+
+One removal, one new module. **The constructor function is `Descriptor.describe`,
+not `Descriptor.of` as the issue writes it — `of` is a Haskell reserved word
+(`case … of`) and cannot be a function name.** `describe` was chosen to read at
+the call site (`Descriptor.describe @AddItem`) alongside the existing nullary
+type-applied primitives `Schema.toSchema @value` and `TypeName.reflect @value`;
+`Descriptor.reflect` and `Descriptor.forType` are the alternatives, and renaming
+is a one-line change at this gate.
+
+The signature also carries **`ToJSON value`**, which the issue's sketch omits.
+It is not optional: `Documented.examples :: Array value` is type-indexed, and
+flattening it to the monomorphic `Array Json.Value` the record holds is exactly
+a `ToJSON` call. Every concept type that reaches a descriptor already has a
+`ToJSON` instance (the TH markers derive it), so the constraint costs existing
+callers nothing.
+
+`Descriptor` is deliberately **not** added to the `Core` re-export list. Its
+field labels (`name`, `description`, `schema`) would collide with the
+`Documented (..)` and `Schema (..)` re-exports already there; it is a
+qualified-import module (`import Descriptor qualified`), and with
+`NoFieldSelectors` its fields are reached through `descriptor.name` anyway. That
+also settles the import-cycle caveat in the issue: `Descriptor` imports
+`Documented`, `Schema` and `Service.Command.Core` directly and never imports
+`Core`, so no cycle can form.
+
+```diff signatures
+- Documented: name :: Documented value => Text
++ Descriptor: data Descriptor
++ Descriptor: Descriptor :: Text -> Text -> Schema -> Array Value -> Bool -> Descriptor
++ Descriptor: [name] :: Descriptor -> Text
++ Descriptor: [description] :: Descriptor -> Text
++ Descriptor: [schema] :: Descriptor -> Schema
++ Descriptor: [examples] :: Descriptor -> Array Value
++ Descriptor: [deprecated] :: Descriptor -> Bool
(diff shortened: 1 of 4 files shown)


NickSeagull:
Superseded: issue #801 re-entered through the new dispatch queue as its first real change (in flight now as bead nh-bwd); this old-pipeline spec PR is closed per the cutover plan. The spec content carries forward into the new molecule's spec step.

Issue #811, opened 2026-08-17
Cleanly integrate beads (bd) into the hix devShell (currently via mkShell.inputsFrom workaround)


NickSeagull:
## Summary

Cleanly integrate `bd` (beads) into the project's `hix`-based Nix devShell. Today it's pulled in via `mkShell.inputsFrom` as a workaround, because `bd`'s package (built with `buildGo126Module`) is incompatible with how `hix` composes its own shell.

## Problem

`hix` generates the project's devShell from the Haskell package set. Adding a Go-built tool (`bd`, built via `buildGo126Module`) directly into that shell hits an incompatibility, so the current workaround pulls `bd`'s derivation in through `mkShell.inputsFrom` rather than as a first-class shell input. This works, but it's fragile — `inputsFrom` composition isn't the intended extension point for "add one more CLI tool to the dev shell", and it may not survive future `hix`/Haskell.nix upgrades cleanly.

## Proposal

Find and apply the proper `hix` extension point for adding a non-Haskell, non-`hix`-managed CLI tool (e.g. `hix`'s `shell.tools` / extra `buildInputs` mechanism, or an explicit `pkgs.mkShell` override that composes cleanly rather than via `inputsFrom`), and migrate `bd`'s devShell integration to it.

## Acceptance criteria

- [ ] `bd` is available in the devShell without relying on `mkShell.inputsFrom`.
- [ ] The integration survives a `hix`/haskell.nix input bump without needing rework (or the fragility is at least reduced/documented).
- [ ] Documented: how to add another non-Haskell CLI tool to the devShell the same way in the future.

PR #822, opened 2026-08-18
spec(007): bd enters the devShell through hix's own extension point (#811)

NickSeagull:
Gate 1 — **spec only**, no implementation. The diff is the contract the
maintainer approves before any code is written.

Closes-spec-for: #811

## What this proposes

Drop the `pkgs.mkShell { inputsFrom = [ flake.devShells.default ]; }` workaround
in `flake.nix` and give `bd` a first-class entry through hix's own extension
point:

1. `flake.nix` — one overlay injecting a single namespaced attribute,
   `neohaskellTools = { bd = beads.packages.${system}.bd; }`, and the `devShells`
   override block deleted.
2. `nix/hix.nix` — `pkgs.neohaskellTools.bd` in `shell.buildInputs` (not
   `shell.tools`, which haskell.nix resolves against the project GHC).
3. `README.md` — the AC3 recipe for adding the next non-Haskell CLI tool.

No version bump, no runtime change: `bd` stays the same `flake.lock`-pinned
derivation on the same `PATH`. `beads.overlays.default` stays unused (it calls
`final.buildGo126Module`, which our pin lacks).

## Contents of this PR

- `docs/changes/007-bd-via-hix-shell-extension-point.md` — the spec: contract
  delta (empty signatures — infrastructure only), criteria C1–C8, failure modes
  F1–F11 with a concurrency section, a Primitives section, an explicit
  edit-set delta, and the limits that are argued rather than tested.
- `docs/decisions/0075-*.md` + its index row and the regenerated ADR landing
  page — why hix's extension point over a cleverer composition, why the
  attribute is namespaced rather than bare, and what happens if hix's module
  `pkgs` turns out not to see our overlay (park as `wrong-localization`, never
  a silent fallback).

## Gates

- `./dev spec-check` — OK (8 specs valid)
- `./dev spec-check --plan` — `design_reviews: []` (no security/perf review
  routed; matches the binding localization)
- `./dev adr-check`, `./dev adr-website --check`, `./dev doctor` — green

## Continue signal

A maintainer `@claude` comment on this PR is the Gate 1 approval.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -37,6 +37,8 @@ jobs:
       # Static workflow wiring: hidden-artifact upload opt-in + retrospect
       # loop stays deterministic/report-only (pure stdlib, no toolchain).
       - run: ./dev workflow-check
+      # Change 007 static proving tests (C1-C3, C8) — no nix needed.
+      - run: ./dev devshell-check --self-test
 
   codemap:
     name: codemap check (localization ontology)
--- a/.github/workflows/test-macos.yml
+++ b/.github/workflows/test-macos.yml
@@ -76,5 +76,8 @@ jobs:
       - name: Build the project with cabal
         run: nix develop --command cabal build all --disable-documentation
 
+      - name: devShell proving tests (change 007 — C4-C6)
+        run: nix develop --command ./dev devshell-check
+
       - name: Run ${{ matrix.test-suite }}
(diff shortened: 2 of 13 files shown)


NickSeagull:
Closing: this spec draft belongs to the retired pre-ADR-0075 process (20-step formula, draft-PR spec gate). The beads state was reset and the change process rebuilt (#837) — if this work is still wanted, it re-enters as a fresh request through the change formula v2.

